PR #1, opened 2020-07-24
add countdown and progress bar

jackyzha0:
image

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -3,18 +3,24 @@ import styled from 'styled-components'
 import GlobalStyle from './theme/GlobalStyle'
 import ThemeProvider from './theme/ThemeProvider'
 
+// temp for testing
+import ProgressBar from './components/ProgressBar'
+import TimeDisplay from './components/TimeDisplay'
+
 const ExampleText = styled.div`
-  color: ${p => p.theme.copy}
+  color: ${p => p.theme.text}
 `
 
 function App() {
   return (
-    <React.Fragment>
+    <>
       <ThemeProvider>
         <GlobalStyle />
         <ExampleText>Live website.</ExampleText>
+        <TimeDisplay days={1} hours={23} minutes={0} seconds={40}></TimeDisplay>
+        <ProgressBar current={33} total={100}></ProgressBar>
       </ThemeProvider>
-    </React.Fragment>
+    </>
   );
 }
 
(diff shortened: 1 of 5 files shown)


src/App.js
```diff
@@ -3,18 +3,24 @@ import styled from 'styled-components'
 import GlobalStyle from './theme/GlobalStyle'
 import ThemeProvider from './theme/ThemeProvider'
 
+// temp for testing
+import ProgressBar from './components/ProgressBar'
+import TimeDisplay from './components/TimeDisplay'
+
 const ExampleText = styled.div`
-  color: ${p => p.theme.copy}
+  color: ${p => p.theme.text}
```
ianmah, line 11: `colors.text`?

src/components/ProgressBar.js
```diff
@@ -0,0 +1,34 @@
...
+  flex: 0 0 ${p => p.percent * 100}%;
+  background-color: ${p => p.theme.colors.highlight};
+`
+
+// rest of progress bar
+const FlexSecondary = styled.div`
+  background-color: ${p => p.theme.colors.primary};
+`
+
+const ProgressBar = ({current, total}) => {
```
ianmah, line 25: do we want to just have progress bar take a percent? Or a value out of 100?

src/components/TimeDisplay.js
```diff
@@ -0,0 +1,60 @@
...
+  width: 6vw;
+  font-size: 4em;
+`
+
+const TimeUnitText = styled(TimeText)`
+  font-size: 1em;
+  margin: 0 10px;
+`
+
+const Seperator = styled.span`
```
ianmah, line 24: nit: `Separator`

jackyzha0: imagine knowing how to spell

src/components/ProgressBar.js
jackyzha0: good idea, that way it would make sense for datetime objects too and we can handle that percentage calculation in the container

src/components/TimeDisplay.js
```diff
@@ -0,0 +1,60 @@
...
+  font-size: 4em;
+`
+
+const TimeFlex = styled.div`
+  display: flex;
+  justify-content: center;
+  align-items: center;
+`
+
+const TimeDisplay = ({ days, hours, minutes, seconds }) => {
```
ianmah, line 35: thoughts on passing a timestamp to count down to here? Or do you want to do that in a container instead?

jackyzha0: Was thinking of doing that in the container

PR #9, opened 2020-07-24
Firebase init

### LINKED ISSUE #7: Connect Firebase

ianmah:
Closes #7


src/utility/firebase.js
```diff
@@ -0,0 +1,18 @@
+import firebase from 'firebase/app'
+import 'firebase/firestore'
+
+if (!firebase.apps.length) {
```
jackyzha0, line 4: What is this conditional checking? To see if there's a firebase app?

src/App.js
```diff
@@ -1,22 +1,31 @@
...
-const ExampleText = styled.div`
-  color: ${p => p.theme.colors.text}
-`
 
 function App() {
+  const [hackathons, setHackathons] = useState()
+
+  // example getting all hackathons data
+  useEffect(() => {
+    db.collection('Hackathons').get().then(querySnapshot => {
```
jackyzha0, line 16: Would it make sense to make a collection for each hackathon so we don't grab more info than we need?

src/App.js
```diff
@@ -1,22 +1,31 @@
...
+      const paths = querySnapshot.docs.map(doc => {
```
jackyzha0, line 17: Why is this var called paths

src/utility/firebase.js
ianmah: Yeah, this file will be imported many times and we only want a single app

src/App.js
ianmah: This is just an example call, we can defs just grab one hackathon's data.

src/utility/firebase.js
jackyzha0: Maybe a good place to use a singleton?

ianmah: this is the singleton

jackyzha0: oh lmao all the singletons i've seen before had a slightly different structure but this works

PR #11, opened 2020-07-31
Common.js and typography section

jackyzha0:
closes #3 and #4

PR #12, opened 2020-08-02
Hacker Countdown Container

jackyzha0:
includes changes from #11 so review that one first


src/containers/HackerCountdown.js
```diff
@@ -0,0 +1,30 @@
...
+
+    return (
+        <>
+            <TimeDisplay days={diff.getDay()} hours={diff.getHours()} minutes={diff.getMinutes()} seconds={diff.getSeconds()}></TimeDisplay>
+            <ProgressBar percent={progress * 100}></ProgressBar>
+        </>
+    );
+}
+
+export default HackerCountdown
```
ianmah, line 30: nit: rename to `Countdown`?

PR #19, opened 2020-09-01
FAQ

### LINKED ISSUE #16: FAQ Section
Database schema will be the same as the existing FAQ section, but will be a different collection.

Search should update as you type. Find good library :)

image

jackyzha0:
Closes #16 

# stuff
* fetches from firestore root FAQ collection (selects current hackathon and livesite specific ones)
* two column responsiveness
* debounced search for faq

# Filter in action
image

# Full thing
image

# Mobile (thin) view
image

### DIFF
--- a/src/components/Accordion.js
+++ b/src/components/Accordion.js
@@ -1,0 +1,26 @@
+import React from 'react';
+import styled from 'styled-components'
+import { H3 } from './Typography';
+
+const Accordion = styled.details`
+  display: inline-block;
+  
+  & > summary:focus {
+    outline: 0;
+  }
+
+  & > summary > * {
+    display: inline-block;
+  }
+`
+
+export default ({ heading, children }) => {
+  return (
+    <Accordion>
+      <summary>
+        <H3>{heading}</H3>
+      </summary>
+      {children}
+    </Accordion>
+  );
+}
--- a/src/components/Accordion.js
+++ b/src/components/Accordion.js
@@ -4,6 +4,8 @@ import { H3 } from './Typography';
 
 const Accordion = styled.details`
   display: inline-block;
+  width: 100%;
+  cursor: pointer;
   
   & > summary:focus {
     outline: 0;
--- a/src/components/Faq.js
+++ b/src/components/Faq.js
@@ -1,12 +1,65 @@
 import React from 'react';
-import { Card } from './Common';
-import { H1, P, H3 } from './Typography';
+import Accordion from '../components/Accordion';
+import { H1, P, H2 } from './Typography';
+import styled from 'styled-components'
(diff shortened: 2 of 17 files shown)


ianmah (commented):
Can we make the FAQ card have a pointer mouse

src/components/Faq.js
```diff
@@ -0,0 +1,82 @@
...
+  & > li {
+    ${CardLike};
+    padding: 0.5em 1em;
+    margin-bottom: 1em;
+    list-style-type: none;
+  }
+`
+
+export default ({ faq }) => {
+  const categories = faq.reduce((q, a) => {
```
ianmah, line 36: is q question and a answer? Can we use longform as variable names instead?

src/components/Faq.js
```diff
@@ -0,0 +1,82 @@
...
+
+  function createFAQList(entries) {
+    return (
+      <DetailFlex>
+        {
+          splitHalf(entries).map(half =>
+            <DetailColumn>{half.map(singleEntry)}</DetailColumn>
+          )
+        }
+      </DetailFlex>
```
ianmah, line 54: is this to make two columns of FAQ boxes?

src/components/Faq.js
```diff
@@ -0,0 +1,82 @@
...
+    );
+  }
+
+  function singleEntry(entry) {
+    return (
+      <li>
+        <Accordion heading={entry.question}>
+          <P>{entry.answer}</P>
```
ianmah, line 62: I think this could use more padding on the bottom

src/components/Accordion.js
```diff
@@ -0,0 +1,26 @@
...
+`
+
+export default ({ heading, children }) => {
+  return (
+    <Accordion>
+      <summary>
+        <H3>{heading}</H3>
+      </summary>
+      {children}
+    </Accordion>
```
ianmah, line 24: the design has the arrow on the right, do you think that violates any best-practices/a11y

src/components/Faq.js
jackyzha0: yes

src/components/Accordion.js
jackyzha0: image

From what I've seen, left-side is generally best practice but I can change it to design if necessary

src/components/Faq.js
ianmah: Any reason why we can't use flexbox wrap instead?

jackyzha0: Then questions might get inadvertently shuffled around once you expand enough Qs, I tried this and the UX is kinda nasty

ianmah (approved):

PR #28, opened 2020-09-19
Deployments

### LINKED ISSUE #24: deployment pipeline
🚀

jackyzha0:
closes #24 

changes
* added GA deployment pipeline
* config firebase hosting
* fixed allowed origins on livestream frame


.github/workflows/deploy.yaml
```diff
@@ -0,0 +1,49 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout Repo
+        uses: actions/checkout@master
+      - name: Install Dependencies
+        run: yarn
+      - name: Build
+        run: yarn build
+        env:
+          REACT_APP_FIREBASE_API_KEY: ${{ secrets.FIREBASE_API_KEY }}
```
ianmah, line 19: just wondering where the secrets are stored

jackyzha0: https://github.com/nwplus/livesite/settings/secrets

ianmah (approved):
LGTM, but I would wait for @Ciuffi to approve before merging

Ciuffi (approved):
this looks good to me! Does it work properly and everything?

Ciuffi:
Nice, have you tested it?

jackyzha0:
Going to update secrets real quick, but is there a way to test the Github action without merging

Ciuffi:
Not really (besides adding the deployment branch to this action, seeing if it runs, then removing it before merging).

Ideally we should know whether the current firebase settings work properly for a deployment. Deploying it locally should give us this information.

.github/workflows/deploy.yaml
```diff
@@ -0,0 +1,49 @@
...
+        uses: actions/checkout@master
+      - name: Download Artifact
+        uses: actions/download-artifact@master
+        with:
+          name: build
+          path: build
+      - name: Deploy to Firebase
+        uses: w9jds/firebase-action@master
+        with:
+          args: firebase deploy --only hosting:production
```
Ciuffi, line 47: Not sure how this action works, but do you need the whole firebase string or just the deploy part?

src/components/Livestream.js
```diff
@@ -55,7 +55,7 @@ const Livestream = () => {
     <Container>
       <TwitchContainer>
         <TwitchFrame
-          src="https://player.twitch.tv/?channel=nwplus&parent=localhost&parent=live.nwplus.io"
+          src="https://player.twitch.tv/?channel=nwplus&parent=localhost&parent=live.nwplus.io&parent=nwplus-ubc.firebaseapp.com"
```
Ciuffi, line 58: Just out of curiosity, Why are you redeclaring the parent query 3 time? Does twitch check each one and take whichever one works, or does it only take the last one?

jackyzha0: Yeah Twitch will just allow all of them as possible origins

.github/workflows/deploy.yaml
jackyzha0: Yeah this action basically just allows you to call the `firebase` CLI from the action so you need the whole string

jackyzha0:
Successfully deployed locally: https://nw-livesite.firebaseapp.com/

.github/workflows/deploy.yaml
Ciuffi: Ah okay, was just a bit confused because the variable there is called args

src/components/Livestream.js
Ciuffi: cool

Ciuffi (approved):
Looks good to me, good job. Now to see if it works 😉

jackyzha0:
🤞

PR #43, opened 2020-10-10
added dynamic sidebar carmen

carmenx:
with help from ian i managed to do some code 

yay


jackyzha0 (commented):
what happens when the src is null? is it just empty image?

ianmah:
@jackyzha0 it will display a broken image :)

PR #45, opened 2020-10-11
Use livesite doc instead of DayOf event for hacking period

ianmah:
- title
- also updated colour and spacing for announcements
image


jackyzha0 (approved):
lgtm 💯

PR #46, opened 2020-10-13
Live Schedule

jackyzha0:
[wip] draft PR

image

* pulls data from firebase
* optimal scheduling with min number of columns
* timeline bar
* add tags for events


src/components/Schedule/Schedule.js
```diff
@@ -0,0 +1,81 @@
...
+          usedIndices.push(i)
+          latestTime = new Date(event.endTime)
+        }
+        return accumulator;
+      }, [])
+
+      // update unused by removing used events
+      unusedEvents = unusedEvents.filter((value, index) => {
+        return usedIndices.indexOf(index) === -1;
+      })
```
ianmah, line 61: I'm guessing theres no way to remove from `unusedEvents` while you are iterating thru it the first time.

Also maybe we can use `.includes` instead of `indexOf` to make the code slightly more intuitive?

src/components/Schedule/Schedule.js
```diff
@@ -0,0 +1,81 @@
...
+      let latestTime = new Date(0) // epoch time lmao
+      let usedIndices = []
+
+      // produce optimal schedule and track which events are used
+      const sched = unusedEvents.reduce((accumulator, event, i) => {
+        const curTime = new Date(event.startTime)
+        if (curTime.getTime() >= latestTime.getTime()) {
+          // set offset as number of hours between start of hackathon and this event
+          const hoursFromStart = msToHours(curTime - hackathonStart)
+          event.hourOffset = hoursFromStart
```
ianmah, line 45: nit: I would think `event.startTime` is more clear but we don't have to change it

src/components/Schedule/Schedule.js
```diff
@@ -0,0 +1,81 @@
...
+  return (
+    <FlexColumn>
+      {column.map((event) => <Event event={event} />)}
+    </FlexColumn>
+  )
+}
+
+export default ({ events, hackathonStart, hackathonEnd }) => {
+  const produceOptimalSchedule = (events) => {
+    const res = []
```
ianmah, line 30: nit: `columns`?

src/components/Schedule/Constants.js
```diff
@@ -0,0 +1,10 @@
+export const HOUR_HEIGHT = 150
+export const EVENT_GAP = 5
+export const EVENT_WIDTH = 400
+
+export const EVENT_TYPES = {
+  "main": ["MAIN", "#FFFFFF"],
```
ianmah, line 6: is `MAIN: "#FFFFFF"` any better? Was thinking accessing indices of an event type was a lil weird. Alternative,
```
main: {
    label: "MAIN",
    colour: "#FFFFFF",
}
```

ianmah (commented):
POG PR ngl, jsut a few nits

src/components/Schedule/Schedule.js
jackyzha0: yeah couldn't figure out how to modify the thing you're reducing on lmao

will include that change though

src/components/Schedule/Constants.js
jackyzha0: I wanted to keep it easy to modify but yeah I agree an object would be nicer to work with

ianmah (approved):
:ship it:

Issue #60, opened 2020-10-31
Sponsor Logo Section


ianmah:
Sponsor logos should be grouped by Tier
Each tier should have a different size

example of grouping by category (tier in this case)
https://github.com/nwplus/livesite/blob/d5bbfd18426cf94cb784ca5e6c4633b573baa58c/src/components/Faq.js#L7-L10

![image](https://user-images.githubusercontent.com/11839441/97770445-93424880-1af0-11eb-96de-2dfb41778609.png)

Issue #52, opened 2020-10-30
Prizes Section (sponsors)


jackyzha0:
This ticket is to build out the components needed for the Sponsors page, along with the logic to fetch the data from Firebase.

## The page should
* fetch sponsor data from Firebase
	* schema: https://github.com/nwplus/database-schemas/blob/ec0f607e838ef3f3753a3318b915a3f3af88b5b5/nw-firestore-schema.ts#L3-L11

## Misc.
Firebase data can be fetched like this.
```js
import { db } from '../utility/firebase'

db
    .collection(DB_COLLECTION)
    .doc(DB_HACKATHON)
    .collection('QuickLinks')
    .get()
    .then(querySnapshot => {
		const data = querySnapshot.docs
		// do something
    });
```

More detailed example implementation here: https://github.com/nwplus/livesite/blob/fdb57f5f5eb37ed37cadc6f65f3b3c6a909daa32/src/containers/Quicklinks.js#L13

## Designs
![image](https://user-images.githubusercontent.com/23178940/97657056-2c585d00-1a26-11eb-97c9-96484a839cb0.png)

acchiang:
sorry for dumb question but looks like firebase only has schema for sponsor data, not sponsored prizes data, is that something we want to add to firebase? 
![image](https://user-images.githubusercontent.com/47487758/98203382-ce75ba80-1ee8-11eb-9035-bee829b71147.png)

and prizes? is that something to add to firebase as well? 

what should happen for HackCamp that doesn't actually have prizes? 

@jackyzha0

jackyzha0:
Yeah, feel free to make a PR to schema repo for that too! I think overall and sponsor prizes would have diff schemas.

HackCamp does have a single sponsor prize so that should be ok, but in the case there isn't, we can display a "No prizes available" message like for the empty FAQ case

acchiang:
@jackyzha0 I made the schema repo PR, I made prizes into one type but main prize vs sponsor prize can be differentiated by its `place` field (similar to how your quicklinks are one type but common vs not common categoried). Lmk if you feel strongly about two separate prize types in the schema! https://github.com/nwplus/database-schemas/pull/15

jackyzha0:
Thoughts on making sponsor prizes a field under a specific sponsor? That way when spocos add new sponsors, they can add sponsor prizes easily without needing to touch livesite stuff. Might make fetching slightly more difficult though.

PR #74, opened 2020-11-05
add stub components to Sponsors.js so people can work on this tab tgt

acchiang:
as title ^ 
so Daniel R and I can work on issues #60 and #52 simultaneously


jackyzha0 (approved):
🧼

PR #76, opened 2020-11-05
Sponsors tab prizes section 

acchiang:
Yes, I know the ordering is kind of messed up for sponsor prizes, can't do anything about it for now because the columns for responsiveness break down in columns first, then rows. Should be fine if we never have 4th 5th 6th place prizes ;) 
image
Screen Shot 2020-11-05 at 12 13 12 AM
Screen Shot 2020-11-05 at 12 13 18 AM

Please note new prize schema in https://github.com/nwplus/database-schemas/pull/15


acchiang:
Very bare if no prizes at all (headers don't show if there are no prizes from firebase): 
image

src/containers/Prizes.js
```diff
@@ -1,17 +1,122 @@
...
+      <UL>
+        {prize.content.split("|").map((item) => <LI key={prize.title + item}>{item}</LI>)}
+      </UL>
+    </PrizeCard>
+  );
+}
 
+
+export const MainPrizes = () => {
+  const [mainPrizes, setMainPrizes] = useState([]);
```
dryu99, line 57: not sure if this would make a diff, but maybe better to maintain prize state in one container component and filter main + sponsor prizes to pass to 2 diff presentational components? That way we wouldn't have to make 2 `getPrizes` db calls every render

src/components/Common.js
```diff
@@ -86,4 +86,46 @@ export const DetailColumn = styled.ul`
...
 export const DetailAnswer = styled(P)`
   margin-bottom: 0.85em;
-`
\ No newline at end of file
+`
+// given array, split into array of chunks of size n 
+// "balanced" (subarrays' lengths differ as less as possible) or 
+// "even" (all subarrays but the last have the same length):
+// https://stackoverflow.com/questions/8188548/splitting-a-js-array-into-n-arrays
+export const chunkify = ((a, n, balanced) => {
```
jackyzha0, line 94: I'd out this in `src/utility` as it's not really a component

acchiang: mmm something like utilities.js? not sure what to call it

src/containers/Prizes.js
acchiang: sounds goodo i'll try this change

acchiang:
Dummy data removed from latest commits but view this link for when the data was still there https://nwplus-ubc--pr76-sponsors-tab-9psn63q8.web.app/sponsors

src/containers/Prizes.js
```diff
@@ -1,15 +1,103 @@
...
+export default () => {
+  const [mainPrizes, setMainPrizes] = useState([])
+  const [sponsorPrizes, setSponsorPrizes] = useState([])
+
+  useEffect(() => {
+    getPrizes()
+      .then(docs => {
+        let prizes = {}
+        prizes.main = Object.values(
+          docs.reduce((result, doc) => {
```
ianmah, line 64: could this be a filter? Same below

src/containers/Prizes.js
```diff
@@ -1,15 +1,103 @@
...
+      })
+      .then(prizes => {
+        setMainPrizes(prizes.main)
+        setSponsorPrizes(prizes.sponsor)
+      })
+  }, [setMainPrizes, setSponsorPrizes])
+
+  let res = []
+  if (mainPrizes.length > 0) {
+    res.push(
```
ianmah, line 87: I've never seen this before, could we use jsx instead? Not sure if this works, maybe something similar?
```
return (
{ mainPrizes.length > 0 && <> ... </> }
{ sponsorPrizes.length > 0 && <> ... </> }
)
```

src/containers/Prizes.js
```diff
@@ -1,15 +1,103 @@
...
   )
 }
+
+export default () => {
+  const [mainPrizes, setMainPrizes] = useState([])
+  const [sponsorPrizes, setSponsorPrizes] = useState([])
+
+  useEffect(() => {
+    getPrizes()
+      .then(docs => {
```
ianmah, line 61: Does react give warnings about this? Something something cannot call async in useEffect? If you aren't seeing console warnings ignore this lol just wanted to check

src/containers/Prizes.js
acchiang: consider it done

src/containers/Prizes.js
acchiang: no errors during development, but good tip to check ;)

src/components/Faq.js
```diff
@@ -2,22 +2,20 @@ import React from 'react'
...
-  const splitHalf = arr => {
-    const half = Math.ceil(arr.length / 2)
-    return [arr.splice(0, half), arr.splice(-half)]
-  }
-
   const createFAQList = entries => {
     return (
       <DetailContainer>
-        {splitHalf(entries).map((half, i) => (
+        {chunkify(entries, COLUMNS_OF_FAQ, true).map((half, i) => (
```
jackyzha0, line 18: 🧼

jackyzha0 (approved):
nice

src/components/Common.js
ianmah: would've preferred `index.js` but im too slow 🐢

PR #77, opened 2020-11-05
Text input component

acchiang:
image


src/components/Common.js
```diff
@@ -44,6 +44,36 @@ export const SecondaryButton = styled(Button)`
   color: ${p => p.theme.colors.primary};
   border: 1px solid ${p => p.theme.colors.primary};
 `
+// todo: error message
+export const TextInput = styled.input.attrs({
+  type: 'text',
+})`
+  ${CardLike};
```
ianmah, line 51: Which parts are we hoping to keep from CardLike?

src/components/Common.js
```diff
@@ -44,6 +44,36 @@ export const SecondaryButton = styled(Button)`
   color: ${p => p.theme.colors.primary};
   border: 1px solid ${p => p.theme.colors.primary};
 `
+// todo: error message
+export const TextInput = styled.input.attrs({
```
ianmah, line 48: Thoughts on making a new file for `Inputs`?

src/components/Common.js
acchiang: not sure yet, i am going to figure out how to implement the error message/state for text inputs

acchiang:
image

acchiang:
WIP: missing hover behaviour for Invalid input

src/components/TextInput.js
```diff
@@ -0,0 +1,72 @@
...
+  }
+  ${p => (p.invalid) &&
+    `border: 2px solid ${p.theme.colors.secondaryWarning};
+    ::placeholder {
+      color: ${p.theme.colors.secondaryWarning};
+    }
+    :hover {
+      border: 2px solid ${p => p.theme.colors.primary};
+    }`
+  }
```
acchiang, line 47: idk why but i can't get `:invalid` to work, lmk if anyone has ideas, this is my hacky solution for now

src/components/TextInput.js
```diff
@@ -0,0 +1,72 @@
...
+  errorMsg
+}) => {
+
+  return (
+    <TextInputContainer>
+      <TextInputBox value={value} onChange={onChange} placeholder={placeholder} disabled={disabled} size={size} invalid={invalid} />
+      {invalid && <ErrorMsg> {errorMsg} </ErrorMsg>}
+    </TextInputContainer>
+  );
+}
```
acchiang, line 70: yes, these props are ugly af, anyone know how to pass props directly into TextInputBox (that's a styled.input)? am dum

src/components/TextInput.js
panjenny0: I think if you change `${p => p.theme.colors.primary}` to `${p.theme.colors.primary}` it should work. 
It's complaining that `${p => p.theme.colors.primary}` is an invalid property value

before
Screen Shot 2020-11-09 at 7 56 06 PM

after
Screen Shot 2020-11-09 at 7 56 55 PM

acchiang: omg @panjenny0 wrinkliest brain in the wild west... thank u

acchiang:
the example text inputs live in /charcuterie

panjenny0 (dismissed):

ianmah:
Does this work for `textarea` components as well?

src/theme/ThemeProvider.js
```diff
@@ -6,7 +6,8 @@ const theme = {
     background: '#2D2937',
     secondaryBackground: '#1D1B24',
     foreground: '#4F4A59',
-    warning: '#E03131',
+    warning: '#EF6C6C',
+    secondaryWarning: '#F18383',
```
ianmah, line 10: I don't really think we need two warning colours @jvssicawu can we just use one?

jvssicawu: yee one is fine for livesite, Im not sure if we use secondary warning for application form though

panjenny0: Seems like the hacker app uses `#F18383` though - should we keep it, or change it to `#EF6C6C` to stay consistent? They are quite close.
Screen Shot 2020-11-15 at 12 04 56 AM

acchiang: i did it because of the _slightly_ different shades here, so i thought we might want two warning colours in general in the future! i'm happy to make the change (to one warning colour), @jvssicawu can you please re-confirm if that's ok? ty!!! 
image

acchiang:
> Does this work for `textarea` components as well?
soz for dumb q, what is a textarea (is it a styled component we already have in the codebase, eg. our search bar in FAQ)?

src/theme/ThemeProvider.js
ianmah: They look identical to me heH

ianmah:
A text area is a larger text box https://developer.mozilla.org/en-US/docs/Web/HTML/Element/textarea

acchiang:
> A textarea is a larger text box https://developer.mozilla.org/en-US/docs/Web/HTML/Element/textarea
> Maybe I should've called it a textarea "element"

Ah gotcha. No there is no textarea implementation yet, I intend to work on it right after this :)  
image

It's not implemented yet because of additional features such as the scroll bar, max char limit. LMK if you have suggestions/preferences on how I should do it (new component? extension? TBD)

src/theme/ThemeProvider.js
acchiang: HAHA same (that's why i added it so late, i realized like last min) 
They're very similar but i checked the figma css and they be different 😢  Will wait on design sign-off but ya i agree i think 1 warning colour is fine

jvssicawu: I'm not sure why there's two, I can check with Tiff but since the difference is so minute i dont think it was intentional. you can change to use just one for now imo

jvssicawu: oh actually LMAO sorry i took a second look because my brain kinda slow at 1am but seems like its specifically for the hover/focus effect so yea its probably intentional actually

ianmah:
I quite like how the Cardlike reusable css works, maybe we could do something similar for the text input and textarea

ianmah (approved):

acchiang:
oops had to merge to get rid of conflicts. pls approve again! thanks :D

panjenny0 (approved):
:shipit:

PR #79, opened 2020-11-06
Sponsor logos

dryu99:
Screen Shot 2020-11-05 at 9 23 08 PM
plat and gold sponsor rows

Screen Shot 2020-11-05 at 9 23 15 PM
silver and bronze sponsor rows

- each row is smaller than the prev by 10%
- hovering over sponsor image will slightly enlarge it (like HackCamp site)
- no inkind sponsors cause we aren't supposed to show them according to sponsorship package (?)
- logos wrap when screen shrinks


acchiang:
quick q how does this look on smaller screens?

src/components/SponsorLogos.js
```diff
@@ -0,0 +1,63 @@
...
+    transform: scale(1.1);
+  }
+`
+
+const platWidth = 250
+const tierData = {
+  platinum: { rank: 1, width: platWidth },
+  gold: { rank: 2, width: platWidth * 0.9 },
+  silver: { rank: 3, width: platWidth * 0.8 },
+  bronze: { rank: 4, width: platWidth * 0.7 },
```
jackyzha0, line 31: is there a way to make this depend on rank? like `width: platWidth - 0.1 * rank`? not sure if that works for objects though

src/containers/SponsorLogos.js
```diff
@@ -1,10 +1,48 @@
-import React from 'react'
+import React, { useEffect, useState } from 'react'
 import styled from 'styled-components'
+import { db } from '../utility/firebase'
+import { DB_COLLECTION, DB_HACKATHON } from '../utility/Constants'
+import Sponsors from '../components/SponsorLogos'
 
-const CenteredH1 = styled.h1`
+const CenteredH1 = styled.h2`
```
jackyzha0, line 7: should be renamed to `CenteredH2`

src/containers/SponsorLogos.js
```diff
@@ -1,10 +1,48 @@
...
+const getSponsors = () => {
+  return db
+    .collection(DB_COLLECTION)
+    .doc(DB_HACKATHON)
+    .collection('Sponsors')
+    .get()
+    .then(querySnapshot => {
+      return querySnapshot.docs
+    })
+}
```
jackyzha0, line 20: if this function is only fetching the firebase stuff, can probably move it to the firebase utils file

src/containers/SponsorLogos.js
```diff
@@ -1,10 +1,48 @@
...
+
+  useEffect(() => {
+    getSponsors().then(docs => {
+      // only keep non-inkind sponsors
+      const filtered = Object.values(
+        docs.reduce((result, doc) => {
+          const data = doc.data()
+          data.tier !== 'inkind' && result.push(data)
+          return result
+        }, [])
```
jackyzha0, line 33: why is a reduce being used here instead of a filter?

src/components/SponsorLogos.js
dryu99: don't think so unless we separate `tierRanks` and `tierWidths` as separate objects. Could put `tierRanks` in `constants.js

dryu99:
> ~quick q how does this look on smaller screens?~
> edit: just checked with the staging link (wow i love that)
> 
> found this while hovering over links, is that a concern?
> Screen Shot 2020-11-05 at 10 05 50 PM

hmm I'm not sure if this can be disabled for `a` tags

ianmah:
> hmm I'm not sure if this can be disabled for `a` tags

I am more concerned about the URL path, looks like it is going to a subdirectory of the livesite instead of going to an external site

dryu99:
> > hmm I'm not sure if this can be disabled for `a` tags
> 
> I am more concerned about the URL path, looks like it is going to a subdirectory of the livesite instead of going to an external site

ah that was my bad when adding link urls in firebase, I didn't write the full urls with `https://`. Should be better now

src/components/SponsorLogos.js
```diff
@@ -0,0 +1,63 @@
...
+  const sortedTiers = Object.keys(tiers).sort((t1, t2) => {
+    return tierRanks[t1] - tierRanks[t2]
+  })
+
+  const createSponsorList = entries => {
+    return <SponsorListContainer>{entries.map(singleSponsor)}</SponsorListContainer>
+  }
+
+  const singleSponsor = (entry, i) => {
+    const imgWidth = platWidth * (1 - 0.1 * tierRanks[entry.tier])
```
jackyzha0, line 50: 🧼

src/components/SponsorLogos.js
```diff
@@ -0,0 +1,63 @@
...
+const SponsorImage = styled.img`
+  max-width: ${p => p.width}px;
+  transition: transform 0.2s;
+
+  &:hover {
+    transform: scale(1.1);
+  }
+`
+
+const platWidth = 250
```
jackyzha0, line 26: nit: `maxSponsorWidth` might be a better name here

jackyzha0 (approved):
single optional comment, but looks spicy 🧞

src/components/SponsorLogos.js
dryu99: agreed

jackyzha0 (approved):
🧼 🧞

PR #94, opened 2020-11-15
hc bRanD reFrEsH

jackyzha0:
* made some colours more consistent
* hc theme yeet

Screen Shot 2020-11-14 at 4 06 38 PM


src/components/Sidebar.js
```diff
@@ -16,10 +17,19 @@ const SidebarContainer = styled.div`
...
+const Logo = styled.img.attrs(p => ({
+  src: p.theme.custom_imgs === 'hc' ? hc_logo : logo,
+}))`
   width: 80px;
   height: 85px;
   margin: 30px 0 0px 50px;
+
+  ${p =>
+    p.theme.custom_imgs === 'hc' &&
+    css`
```
ianmah, line 29: why do we need this?

jackyzha0: HC logo is more rectangle than square, so just fixing width and padding to make it look nicer

ianmah: You can't do that without `css`?

jackyzha0: how else should i do the conditional css bit?

ianmah: does this not work
```
  ${p =>
    p.theme.custom_imgs === 'hc' &&
    `
      width: 120px;
      margin: 30px 0 0px 60px;
    `}
`
```

jackyzha0: lemme try that

src/components/Common.js
```diff
@@ -37,6 +38,14 @@ export const Button = styled.a`
   padding: 0.75em 2.5em;
   border-radius: 3px;
   margin: 1em;
+
+  ${p =>
+    p.theme.custom_imgs === 'hc' &&
+    css`
```
ianmah, line 44: same thing here?

ianmah (approved):

src/components/Common.js
jackyzha0: this file already imports css so i think it should be fine lol

PR #99, opened 2020-11-16
Theme switcher

ianmah:
gif

### DIFF
--- a/src/components/Common.js
+++ b/src/components/Common.js
@@ -1,4 +1,5 @@
-import styled, { css } from 'styled-components'
+import React from 'react'
+import styled from 'styled-components'
 import { text } from './Typography'
 import { P } from './Typography'
 import buttonBG from '../assets/hc_button.svg'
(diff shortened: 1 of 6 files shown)


src/components/Common.js
```diff
@@ -41,13 +42,19 @@ export const Button = styled.a`
...
     p.theme.custom_imgs === 'hc' &&
-    css`
+    `
       background: url(${buttonBG});
       background-size: contain;
       background-repeat: no-repeat;
     `}
 `
 
+export const Button = props => (
```
jackyzha0, line 52: what is this change doing exactly?

jackyzha0 (approved):

src/components/Common.js
ianmah: Buttons without `href` do not change pointer to cursor and do not appear clickable/have good keyboard accessibility. This change is to add a default `href`

jackyzha0: the more you know! thanks

ianmah: Actually this is an `a` element. Not exactly sure what happens if it were a `button` elem but what i said is defs true for `a`s

PR #102, opened 2020-11-17
Dropdown component

andykwan1124:
Used the react-select library to make the dropdown component. Lowkey regret not making one from scratch smh. 

Doc links: 
Styling -> [https://react-select.com/styles](url) -> scroll all the way to the bottom for styled components
Custom Components -> [https://react-select.com/styles](url)
Props -> [https://react-select.com/props](url)


src/theme/ThemeProvider.js
```diff
@@ -49,6 +58,7 @@ const hackcampTheme = {
     text: '#fff',
     link: '#fff',
     linkHover: '#FFBC96',
+    dropdown,
   },
   typography: {
     h1: {
```
acchiang, line 64: nwTheme and HackCamp theme have different colours, could you make two versions here? ty! 

for the HackCamp colours, you can consult Sandy Co (HackCamp design lead) on Slack, @jackyzha0 (HackCamp director), or @jvssicawu (livesite design lead)

src/components/Dropdown.js
```diff
@@ -0,0 +1,270 @@
...
+      }}
+      maxMenuHeight={200}
+      emptySearchDefaultOption={emptySearchDefaultOption}
+      errorMessage={errorMessage}
+      canCreateNewOption
+      isValid={isValid}
+      {...userProps}
+    />
+  ) : (
+    <StyledNormalDropdown
```
acchiang, line 251: can we make the search function debounced (ie. uses a setTimeout to throttle searching)? would be a nice-to-have for the dropdowns that have a loooot of data. 

the searchbar in the FAQ page uses debounced search, feel free to borrow the logic there

acchiang (commented):
will leave the approval to the real devs here @ianmah @jackyzha0 @Ciuffi @panjenny0 but i left a few comments + i just wanted to say I checked out the dropdowns in your staging /charcuterie and it's v nice (v robust options), good job!! 

minor nitpick, is it possible to get the selection highlight to extend under the scrollbar like in the designs? if not we should get @jvssicawu / designers to sign off on the difference (nbd imo but they're the pros)
image
design: 
image

src/theme/ThemeProvider.js
```diff
@@ -1,6 +1,14 @@
 import React from 'react'
 import { ThemeProvider } from 'styled-components'
 
+const dropdown = {
+  background: '#06C1C0',
+  hover: '#D9FFF9',
+  neutral: '#BEBEBE',
+  scrollbar: '#615b82',
+  error: '#F18383',
+}
```
ianmah, line 10: Ideally these should re-use theme colours. 

maybe
`dropdown.background` -> `primary`
`dropdown.hover` -> make a new theme colour for this one I think
`dropdown.neutral` -> default
`dropdown.scrollbar` -> `foreground`
`dropdown.error` -> `secondaryWarning`

src/components/Dropdown.js
```diff
@@ -0,0 +1,270 @@
...
+      width="10"
+      height="8"
+      viewBox="0 0 10 8"
+      xmlns="http://www.w3.org/2000/svg"
+    >
+      <path d="M4.53582 6.89842C4.93534 7.44394 5.74984 7.44394 6.14936 6.89842L9.67032 2.09087C10.1541 1.43027 9.68237 0.5 8.86355 0.5H1.82163C1.00282 0.5 0.531054 1.43027 1.01486 2.09086L4.53582 6.89842Z" />
+    </svg>
+  )
+
+const StyledDropdown = styled(DropdownIcon)`
```
ianmah, line 139: maybe
```suggestion
const StyledDropdownIcon = styled(DropdownIcon)`
```

src/components/Dropdown.js
```diff
@@ -0,0 +1,270 @@
...
+  .react-select__option {
+    color: ${p => p.theme.colors.background};
+    overflow-wrap: break-word;
+  }
+
+  .react-select__menu {
+    background-color: ${p => p.theme.colors.dropdown.background};
+    border-radius: 7px;
+    margin-top: 0px;
+    max-width: ${p => (p.isSearchable ? `387px` : `287px`)};
```
panjenny0, line 39: I know you followed the exact dimensions of the components on figma, but since the no-search dropdown width should match that of the inputs, I think you could do something similar to what @acchiang did for the textinputs: 
```
const inputSize = {
  default: '300px',
  medium: '400px',
  large: '600px',
}
```
and the set max-width to something like
```
max-width: ${p => (p.isSearchable ? inputSize.medium : inputSize.default)};
```
Same with line 15.

Screen Shot 2020-11-21 at 12 03 30 AM

panjenny0 (changes requested):
Great job!! This is very well done and it seems like you put a lot of work into it  🙌  
One thing I would add is the secondary error state when you hover over an invalid dropdown. 
#### no hover
Screen Shot 2020-11-21 at 12 10 44 AM

#### hover 
Screen Shot 2020-11-20 at 11 47 41 PM

The behaviour should be similar to TextInput where we use `theme.colors.secondaryWarning` on hover (`secondaryWarning` looks similar to `secondaryWarning` but there's a slight noticeable difference)

#### no hover
Screen Shot 2020-11-21 at 12 10 04 AM

#### hover
Screen Shot 2020-11-21 at 12 10 20 AM

Regarding making the scrollbar overlap the options on hover, it might not be super straightforward - I tried this but it didn't work. I wouldn't spend over an hour on it in the interest of time.
https://stackoverflow.com/questions/39253887/how-to-make-scrollbar-overlapping-page-content

andykwan1124:
Thanks for the feedback guys! Sorry for the late reply, been busy for the week. @acchiang @panjenny0 I also spent a long time trying to offset the scrollbar from the edge of the container, but I just can't get it to work. I will see what I can do tmrw but it prob won't be my priority :)

src/theme/ThemeProvider.js
andykwan1124: @ianmah would using `linkHover` be fine for the hover state?

ianmah: I think so, just double check with design if you're unsure

src/theme/ThemeProvider.js
andykwan1124: Resolved below in Ian's comment

andykwan1124:
Scrollbar and menu item on-hover colors are awaiting design's decision, but imo the current one looks pre decent? 👀

panjenny0 (approved):
looks good! I think the spacing with the scrollbar is fine - it is indeed very hard to find a workaround and I've looked into it as well. I think it's ok to merge, if anything, we can come back to address it in a future PR when we have time at the end (Just don't want us to be blocked for too long). I'll open an issue if design wants to keep the scrollbar overlapping the hover highlight.

src/components/Dropdown.js
```diff
@@ -0,0 +1,300 @@
...
+
+const dropdownWidth = {
+  normal: '300px',
+  searchable: '400px',
+}
+
+const sharedStyle = css`
+  .react-select__control {
+    background-color: transparent;
+    margin: 2em 0;
```
panjenny0, line 18: I took another look - can we decrease the top margin and increase the bottom? Also you can refer to my other comment below for the additional bottom margin
```suggestion
    margin: 1em 0 2.5em 0;
```

src/components/Dropdown.js
```diff
@@ -0,0 +1,300 @@
...
+  font-weight: bold;
+  padding: 8px 12px;
+`
+
+const StyledErrorMsg = styled.div`
+  color: ${p => p.theme.colors.warning};
+  font-weight: 400;
+  font-size: 16px;
+  padding: 8px 0px;
+  margin-top: -30px;
```
panjenny0, line 114:
```suggestion
  position: absolute;
  top: 56px;
```
I realized that the error message pushes the subsequent content downwards so the content underneath moves up when you click on the dropdown. I think you can make it fixed by having `postion: absolute`

#### before
ezgif com-gif-maker (2)

#### after
ezgif com-gif-maker (1)

panjenny0 (changes requested):
Took another look - can you make one minor tweak to the margin and the error message? Thanks!

panjenny0 (approved):

PR #103, opened 2020-11-17
TextArea (paragraph text input) component

acchiang:
Designs: 
image
Charcuterie (HackCamp theme): 
image
image
image
Note that the scrollbar does not have margin between it and the side of the TextArea like the designs: unfortunately this is not easily doable because webkit-scrollbar does not support padding/margin (and neither does its textarea container in terms of the scrollbar).

### DIFF
--- a/src/components/TextArea.js
+++ b/src/components/TextArea.js
@@ -19,7 +19,7 @@ const TextAreaBox = styled.textarea.attrs({
   ::-webkit-scrollbar-thumb {
     background-color: transparent;
     border-radius: 10px;
-    border: 2px solid ${p => p.theme.colors.highlight};
+    border: 1px solid ${p => p.theme.colors.highlight};
   }
   ::-webkit-scrollbar-track {
     background-color: transparent;
--- a/src/components/TextArea.js
+++ b/src/components/TextArea.js
@@ -1,0 +1,79 @@
+import React, { useEffect, useState } from 'react'
+import styled from 'styled-components'
+import { TextInputLike } from '../components/Common.js'
+
+const TextAreaContainer = styled.div`
+  margin: 1em;
+`
+
+const TextAreaBox = styled.textarea.attrs({
+  type: 'text',
+})`
+  height: 200px;
+  width: 600px;
+  ${TextInputLike};
+  ::-webkit-scrollbar {
+    width: 10px;
+    height: 10px;
+  }
+  ::-webkit-scrollbar-thumb {
+    background-color: transparent;
+    border-radius: 10px;
+    border: 2px solid ${p => p.theme.colors.highlight};
+  }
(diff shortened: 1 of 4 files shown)


acchiang:
ezgif com-gif-maker

ianmah:
Do you mind moving the text input elements outside of the card on the charcuterie page? Looks like in figma they are never actually used inside a card (the card colour makes the text not very accessible)

acchiang:
bump @ianmah :bulbaowo:

src/components/TextArea.js
```diff
@@ -0,0 +1,79 @@
...
+    }
+  }, [maxLength, value])
+
+  return (
+    <TextAreaContainer>
+      <TextAreaBox
+        value={value}
+        onChange={val => {
+          onChangeValue(val.target.value)
+        }}
```
panjenny0, line 66: Just to keep the `onChange` handler consistent with all of the other input components that have been built (text, checkbox, radio, etc), can we change this to 
```suggestion
        onChange={onChange}
```
and rename `onChangeValue` to `onChange` in line 41?

acchiang: thanks for pointing this out! will follow up after I make a change (that's not exactly the same as this suggestion)

src/components/TextArea.js
```diff
@@ -61,9 +61,7 @@ export const TextArea = ({
     <TextAreaContainer>
       <TextAreaBox
         value={value}
-        onChange={val => {
-          onChangeValue(val.target.value)
-        }}
+        onChange={val => onChange(val.target.value)}
         invalid={invalid || isLengthExceeded}
```
acchiang: @panjenny0 soo this isn't _exactly_ what you suggested, but I think I captured it in essence (ie. the element now uses `onChange`) 

The reason for `val.target.value` in the arrow function here is because <textarea>'s native onChange callback passes in an object (not a text value), so we have to use `val.target.value` in order to access the value string inside the <textarea> element

otherwise we get this haha
ezgif com-gif-maker (2)

anyway, minor difference imo, hope this is ok to pass review! :D

src/components/TextArea.js
```diff
@@ -0,0 +1,77 @@
...
+  ::-webkit-scrollbar-corner {
+    background-color: transparent;
+  }
+  ::-webkit-resizer {
+    background-color: transparent;
+  }
+`
+const ErrorMsg = styled.p`
+  color: ${p => p.theme.colors.warning};
+  margin: 7px 0px;
```
ianmah, line 36: `em`?

acchiang: oh sure (i got the px from figma mb)

acchiang: image

ianmah (approved):

PR #106, opened 2020-11-17
set toggle colours depending on checked status

acchiang:
HackCamp:
ezgif com-gif-maker (1)

nwTheme: 
image
image

### DIFF
--- a/src/components/ToggleSwitch.js
+++ b/src/components/ToggleSwitch.js
@@ -46,7 +46,7 @@ const ToggleSwitchGraphic = styled.div`
 
   ${p =>
     p.checked &&
-    css`
+    `
       background: ${p => p.theme.colors.foreground};
     `};
 `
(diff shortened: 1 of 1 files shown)


src/components/ToggleSwitch.js
```diff
@@ -43,6 +43,12 @@ const ToggleSwitchGraphic = styled.div`
       cursor: not-allowed;
       opacity: 0.3;
     `};
+
+  ${p =>
+    p.checked &&
+    css`
```
ianmah, line 49:
```suggestion
    `
```
I don't think we need css above either

ianmah (approved):
LGTM, if you'd like to I think we can totally remove the `css` important and its usages from this file but that's not really your change so feel free to merge without changing that

acchiang:
> LGTM, if you'd like to I think we can totally remove the `css` important and its usages from this file but that's not really your change so feel free to merge without changing that

mm ya i won't do it in this PR but will look into it

PR #119, opened 2020-11-22
Button components

acchiang:
image


src/components/JudgingCard.js
```diff
@@ -54,7 +55,9 @@ export default ({ imgUrl, title, teamName, description, href = '#!' }) => {
         <Title>{title}</Title>
         <TeamName>by {teamName}</TeamName>
         <P>{cutString(description, MAX_CHARACTERS_IN_DESCRIPTION)}</P>
-        <StyledButton href={href}>Judge this submission</StyledButton>
+        <StyledButton color="tertiary" onClick={() => window.open(href, '_blank')}>
```
ianmah, line 58: I thought we are making this link to a page not to a new tab

src/containers/Quicklinks.js
```diff
@@ -42,7 +42,12 @@ export const CommonLinks = () => {
   return (
     <ButtonContainer>
       {links.map(link => (
-        <Button key={link.href} href={link.href} rel="noopener noreferrer" target="_blank">
+        <Button
+          key={link.href}
+          onClick={() => window.open(link.href, '_blank')}
```
ianmah, line 47: Any reason we are switching to onClick instead of href?

src/pages/Charcuterie.js
```diff
@@ -23,7 +24,9 @@ export default () => {
 
   return (
     <>
-      <SecondaryButton href={`javascript:(${toggleTheme})()`}>Toggle Theme</SecondaryButton>
+      <Button color="secondary" onClick={toggleTheme}>
```
ianmah, line 27: We need to keep the javascript: thingy so that its useable as a bookmarklet

src/pages/Charcuterie.js
```diff
@@ -53,8 +56,18 @@ export default () => {
       <Card>
         <H2>Card Element</H2>
         <P>It can contain content. And even buttons!</P>
-        <Button>Primary</Button>
-        <SecondaryButton>Secondary</SecondaryButton>
+        <Button color="primary">Primary</Button>
```
ianmah, line 59: sometimes i think constants are good and sometimes I think man we shouldve just used strings. Wondering if you considered using constants for the colors? Not sure what I'd prefer tbh

src/components/Button.js
```diff
@@ -0,0 +1,87 @@
...
+  }
+  :focus {
+    outline: 0;
+  }
+  :disabled {
+    cursor: not-allowed;
+    opacity: ${p => p.theme.opacity.disabled};
+  }
+  ${p =>
+    (p.color === buttonColors[0] || !buttonColors.includes(p.color)) && // primary color (gradient button) or not specified color
```
ianmah, line 37: maybe? (see constants comment below)
```suggestion
    (p.color === "primary" || !p.color) && // primary color (gradient button) or not specified color
```

src/components/Button.js
```diff
@@ -0,0 +1,87 @@
...
+  default: '80px',
+  large: '300px',
+}
+
+const buttonColors = ['primary', 'secondary', 'tertiary']
+
+const StyledButton = styled.button`
+  display: block;
+  text-decoration: none;
+  font-family: 'HK Grotesk';
```
ianmah, line 17: Maybe we should add this to the themeprovider

src/components/Button.js
```diff
@@ -0,0 +1,87 @@
...
+  transition: all 250ms;
+  min-width: ${p => (p.size ? buttonSize[p.size] : buttonSize['default'])};
+  text-align: center;
+  padding: 0.75em;
+  border-radius: 3px;
+  margin: 1em;
+  :hover {
+    cursor: pointer;
+  }
+  :focus {
```
ianmah, line 29: I'd almost want to remove this - because if we add new variants we have to remember to include a focus, which we've already forgotten for the tertiary variant (no visible change on focus to tertiary button, not good for keyboard accessibility)

src/components/Button.js
```diff
@@ -0,0 +1,87 @@
...
+    `color: ${p.theme.colors.background};
+    background: ${p.theme.colors.primaryGradient};
+    :hover {
+      background: ${p.theme.colors.primaryGradientHover};
+    }
+    :disabled {
+      background: ${p.theme.colors.primaryGradient};
+    }
+    ${
+      p.theme.custom_imgs === 'hc' &&
```
ianmah, line 47: hmM I know you didn't create this but maybe we should change this to `themeName` instead of `custom_imgs`, since we aren't really using it for an image but we are using it to check which theme we are on.
If we actually want to use `custom_imgs` maybe we should export the url to the asset from the themeprovider instead of importing the asset. But only button uses the background image so i dont think its a big deal.

ianmah (commented):
Awesome stuff. Also was wondering why the buttons look smaller in this PR? I don't see any explicit changes to the font-size, was this on purpose? 🤔

acchiang:
good review @ianmah ! kinda swamped for the next two days but i'll get back to u on them asap

ianmah:
Oh my bad didnt realize this is WIP. (You can set PRs to a draft https://github.blog/2019-02-14-introducing-draft-pull-requests/)

ianmah:
Let's make sure to check with design before this is merged if the new button sizes are correct, they seem small

acchiang:
> Oh my bad didnt realize this is WIP. (You can set PRs to a draft https://github.blog/2019-02-14-introducing-draft-pull-requests/)

OH TIL!!! ty, i didn't use github in any of my past internships and was confuse

+ sounds g on the design thing!

src/pages/Charcuterie.js
acchiang: following up, i did actually use constants in Button.js for these strings, i'll pull those out as an import so we can use them here 
 `const buttonColors = ['primary', 'secondary', 'tertiary']`

src/components/Button.js
acchiang: done!

acchiang: `font-family: ${p => p.theme.font};`

src/components/Button.js
acchiang: oop yes update: got rid of constants cuz primary secondary are clear enough

src/pages/Charcuterie.js
acchiang: yep we're just using strings now (i checked some libraries and ya consts here just complicate things)

src/components/Button.js
acchiang: focus has been added! it's different than designs so i'll be touching base w. them

src/components/Button.js
acchiang: i think i'll leave this for now bc it's @jackyzha0's work here and i'm not 100% sure what the expectations for themes would be, going forward? is hc the only event with picture buttons?

acchiang:
sorry moved back to draft bc i forgot to talk to design

acchiang:
no update from design yet, i'm going ahead with dev review because this blocks hacker app work, and all the design review parts will come later (possibly in a new issue if needed)

src/components/Button.js
```diff
@@ -0,0 +1,117 @@
...
+  font-family: ${p => p.theme};
+  font-weight: ${p => p.theme.typography.h2.weight};
+  border: transparent;
+  transition: all 250ms;
+  width: ${p => (p.width ? buttonWidth[p.width] : buttonWidth['default'])};
+  text-align: center;
+  padding: ${p => (p.height ? buttonHeightPadding[p.height] : buttonHeightPadding['default'])}
+    0.75em;
+  border-radius: 3px;
+  margin: 1em;
```
ianmah, line 31: I think this can be decided by user? And left out?

ianmah (commented):
Alignment looks off. Are we supporting an icon in the button, like in the designs? If so could you add examples with and without icons in /Charcuterie pls. If you haven't already added support for the icons I say dwbi for now, we should focus on other things

image

EDIT: Also were you planning on implementing this hover:
image

src/components/Button.js
acchiang: do you mean the margin?

ianmah: Yea maybe remove margin on the default buttons

acchiang:
> Alignment looks off. Are we supporting an icon in the button, like in the designs? If so could you add examples with and without icons in /Charcuterie pls. If you haven't already added support for the icons I say dwbi for now, we should focus on other things
> 
> image

oop this is a regression bug i didn't catch in my last commit fixing the size to look more like the figma mockups, i'll look into it 

the (centralized) button designs didn't have icons, but i just noticed other places in the figma that have icons added in. awkward, mb. 

no we don't have defined support for icons rn

src/components/Button.js
acchiang: the padding around the text here makes the button height (also helps the text inside stay vertically centered)

acchiang: image
lmao

acchiang: all g i added margin to those but removed it from Button.js

panjenny0 (changes requested):
Looks solid! Although I still think that a margin should be included. I could see consumers potentially having to add additional styling for the buttons to adjust its alignment with the other components like input/textarea

For example, they might have to add an additional container for the button for these two instances
Screen Shot 2020-11-25 at 6 15 47 PM
Screen Shot 2020-11-25 at 6 19 48 PM

Screen Shot 2020-11-25 at 6 15 56 PM

acchiang:
margins have been reverted now 🤣

panjenny0 (approved):
very nice! 👍

ianmah (approved):
Would be nice to mitigate the image load delay when hovering the button but otherwise 👍

src/pages/Charcuterie.js
```diff
@@ -80,8 +83,39 @@ export default () => {
       <Card>
         <H2>Card Element</H2>
         <P>It can contain content. And even buttons!</P>
-        <Button>Primary</Button>
-        <SecondaryButton>Secondary</SecondaryButton>
+        <P>Colors</P>
+        <Button color="primary">Primary</Button>
+        <Button color="primary" disabled={true}>
```
ianmah, line 88:
```suggestion
        <Button color="primary" disabled>
```

acchiang:
please go ahead and merge once you approve @ianmah @panjenny0

panjenny0 (approved):

PR #136, opened 2020-11-28
124- Resume upload component

### LINKED ISSUE #124: Resume upload
Resume upload component for question 12 in part 2 of the application
(does not include upload to firebase)
Screen Shot 2020-11-24 at 6 51 56 PM

vincentchiang1:
Closes #124 
Implemented front end component for resume button upload. 

Some notes/questions:
- currently using the existing button component, but looks like it doesn't _exactly_ match the Figma, so not sure if I should make a new style?
- right now i'm just displaying the warning text "Please upload..." during the onChange event when **no file** is uploaded for visual purposes only. I'm assuming the alert is intended to be displayed when the user does not upload a file and tries to move to the next page clicking the "Next" button at the bottom as seen on the Figma, but was not sure if I should be making that change here in this issue.

Open to feedback and improvements, please let me know!

Screencast attached: https://drive.google.com/file/d/1PVuw5bUlKW_zPbvrWILFh1bcFZQzc2rE/view?usp=sharing
Screenshots:
image
image


src/components/ResumeUploadBtn.js
```diff
@@ -0,0 +1,41 @@
+import React, { useRef, useState } from 'react'
+import { Button } from '../components/Button'
+import { QuestionHeading } from '../components/Typography'
+
+const resumeStyle = {
+  display: 'flex',
+  alignItems: 'center'
+};
```
ianmah, line 8: We use styled components instead of style objects. This can be changed to 
```suggestion
const ResumeContainer = styled.div`
  display: flex;
  alignItems: center;
`
```

src/components/ResumeUploadBtn.js
```diff
@@ -0,0 +1,41 @@
...
+
+const ResumeUploadBtn = () => {
+  const [hint, setHint] = useState(null)
+  const inputFile = useRef(null);
```
ianmah, line 12:
```suggestion
  const [hint, setHint] = useState()
  const inputFile = useRef()
```

src/components/ResumeUploadBtn.js
```diff
@@ -0,0 +1,41 @@
...
+      setHint(<span>{e.target.value.match(/[\/\\]([\w\d\s\.\-\(\)]+)$/)[1]}</span>)
+    }
+
+    else {
+      setHint(<span style={{ color: '#F18383' }}>Please upload your resume</span>)
+    }
+  }
+
+  var resumeFile = <input ref={inputFile} type='file' hidden='hidden' onChange={handleChange} />
+  var button = <Button style={{ marginLeft: '9em' }} color="secondary" onClick={handleClick}>Upload</Button>
```
ianmah, line 29: These can be `const`

Also, to style a component, create a styled component
```
const StyledButton = styled(Button)`
    margin-left: 9em;
`
```
and then use the StyledButton component instead of the Button component

src/components/ResumeUploadBtn.js
```diff
@@ -0,0 +1,41 @@
...
+    if (e.target.value) {
+      setHint(<span>{e.target.value.match(/[\/\\]([\w\d\s\.\-\(\)]+)$/)[1]}</span>)
+    }
+
+    else {
+      setHint(<span style={{ color: '#F18383' }}>Please upload your resume</span>)
+    }
+  }
+
+  var resumeFile = <input ref={inputFile} type='file' hidden='hidden' onChange={handleChange} />
```
ianmah, line 28: Could we do this?
```suggestion
  var resumeFile = <input ref={inputFile} type='file' hidden onChange={handleChange} />
```

src/components/ResumeUploadBtn.js
```diff
@@ -0,0 +1,41 @@
+import React, { useRef, useState } from 'react'
+import { Button } from '../components/Button'
+import { QuestionHeading } from '../components/Typography'
+
+const resumeStyle = {
+  display: 'flex',
+  alignItems: 'center'
+};
+
+const ResumeUploadBtn = () => {
```
ianmah, line 10:
```suggestion
export default () => {
```

src/components/ResumeUploadBtn.js
```diff
@@ -0,0 +1,41 @@
...
+    inputFile.current.click();
+  }
+
+  const handleChange = (e) => {
+    if (e.target.value) {
+      setHint(<span>{e.target.value.match(/[\/\\]([\w\d\s\.\-\(\)]+)$/)[1]}</span>)
+    }
+
+    else {
+      setHint(<span style={{ color: '#F18383' }}>Please upload your resume</span>)
```
ianmah, line 24: Should be using theme colours here. Here's an example of how to do that:
https://github.com/nwplus/livesite/blob/8a746ca40a131660faf90761ea5890f35706c05f/src/components/Common.js#L110-L113

src/components/ResumeUploadBtn.js
```diff
@@ -0,0 +1,41 @@
...
+
+const ResumeUploadBtn = () => {
+  const [hint, setHint] = useState(null)
+  const inputFile = useRef(null);
+
+  const handleClick = () => {
+    inputFile.current.click();
+  }
+
+  const handleChange = (e) => {
```
ianmah, line 18: Maybe we shold make this function call a callback function? We are using the container pattern in this repo, aka display components shouldn't handle logic, so we'd probably have a wrapper container handle the file once it's added.

ianmah:
Good stuff! Mostly just `styled-components` stuff to change

src/components/ResumeUploadBtn.js
panjenny0: Also can we move the span in hint to the `return ()` instead of including it with hint?

src/pages/Charcuterie.js
```diff
@@ -279,6 +280,11 @@ export default () => {
         debounceEnabled
         throttleTime={1000}
       />
+      <QuestionHeading>question 12</QuestionHeading>
+      <div>
+        <ResumeUploadBtn />
```
panjenny0, line 285: I think having the `marginLeft: '9em'` styling here on the parent div would be more appropriate to show how it's used and how you could style it
With the suggestions I have for passing in additional props, I think keeping track of hint and handleChange could be moved here
something like
```
<ResumeUploadBtn 
     hint={hint.match(/[\/\\]([\w\d\s\.\-\(\)]+)$/)[1]} 
     onChange={(e) => setHint(e.target.value)} 
     isValid={hint} 
     errorMsg="Please upload your resume" 
/>
```
and at the top of this file you would have `const [hint, setHint] = ...` etc

panjenny0 (commented):
Good work!! Left some comments to separate _how_ the component is being rendered with _what_ is being rendered

src/components/ResumeUploadBtn.js
vincentchiang1: thanks for the comment, but it seems to behave differently now and the elements are no longer aligned if you have any ideas as to the cause? I found one person with a similar issue when googling but the fix was not applicable :(
image

src/components/ResumeUploadBtn.js
vincentchiang1: Thanks for the comments! I've moved the spans to the return so that the hint is only string values. I've also replaced the span for the error mssg to use the TextInputLikeErrorMsg component! Is there another component that I should be using for the non-error message (ie. the resume file name) instead of span?

src/components/ResumeUploadBtn.js
ianmah: Could you push up what you changed so I can take a look?

ianmah: Oh. It should be `align-items` not `alignItems`

vincentchiang1: oooooh ok that fixed it thank you boss. and i'll try to push by tonight. almost done addressing all the comments here just tryna fix stuff up!

src/pages/Charcuterie.js
vincentchiang1: thanks for the suggestion! I've made the change to pass the props and moved the hint state into this file as well. But I'm not entirely sure what you mean by moving the marginLeft styling here since originally i used that styling so that there is more space between the "Resume" label on the left and the button + hint on the right. I'm not too sure how I would use the margin here in this case?

src/components/ResumeUploadBtn.js
vincentchiang1: resolved in latest commit!

src/components/ResumeUploadBtn.js
vincentchiang1: thanks for the comments! resolved in latest commit!

src/components/ResumeUploadBtn.js
vincentchiang1: thanks for the comments! resolved in latest commit!

src/components/ResumeUploadBtn.js
vincentchiang1: thanks for the comments! resolved in latest commit!

src/components/ResumeUploadBtn.js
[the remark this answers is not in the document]
vincentchiang1: thanks for the comments! resolved in latest commit!

src/components/ResumeUploadBtn.js
```diff
@@ -0,0 +1,41 @@
...
+
+  var resumeFile = <input ref={inputFile} type='file' hidden='hidden' onChange={handleChange} />
+  var button = <Button style={{ marginLeft: '9em' }} color="secondary" onClick={handleClick}>Upload</Button>
+
+  return (
+    <div style={resumeStyle}>
+      <QuestionHeading>resume</QuestionHeading>
+      {resumeFile}
+      {button}
+      {hint}
```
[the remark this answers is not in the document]
vincentchiang1: thanks for the comments! resolved in latest commit!

src/pages/Charcuterie.js
vincentchiang1: also I wasn't sure how to do the regular expression matching here. since it's possible for the hint to be undefined/null still if the state is undefined initially. I tried it with an initial state of empty string, but then the match result would be null I believe which also causes an error when trying to access the index [1] and I tried using conditionals to only call the match if the hint exists, but it seems like the hint state does not update properly? so I couldn't get it to work and decided to leave the matching in the Upload component still, let me know if that's alright!

src/components/ResumeUploadBtn.js
```diff
@@ -0,0 +1,38 @@
...
+  const handleClick = () => {
+    inputFile.current.click()
+  }
+
+  const resumeFile = <input ref={inputFile} type="file" hidden onChange={onChange} />
+  const button = (
+    <Button color="tertiary" onClick={handleClick}>
+      Upload
+    </Button>
+  )
```
ianmah, line 24: It's not best practice to define components inside other components like this. I'm not 100% sure if the concept applies to functional components but it's better to get in the habit of not doing it.

I believe it has something to do with recreating new child components every time the component re-renders

Just move these components above line 12, and pass in the things u need like inputFile, onChange etc as props

vincentchiang1: oooh makes sense thank you. i think i'll wait for Jenny to comment on any other suggestions first and I'll just make all those changes together later for the next commit

vincentchiang1: ok i'm kinda confused LMAO uhh i'm not super clear on what you mean by "pass in the things u need like inputFile, onChange etc as props" and how to do that hehe. so if i wanna move resumeFile and button outside of the component to the top of the class, how do I access the onChange prop as well as the handleClick, inputFile which are defined inside the component?

ianmah:
```
const ResumeFile = ({inputFile, onChange}) => {...}

// inside default export
<ResumeFile inputFile={inputFile} onChange={onChange} />
```

vincentchiang1: omg whatttt ok that's cool didn't think about that big thanks sir

panjenny0 (approved):

panjenny0 (approved):

Issue #84, opened 2020-11-08
Progress bar + form content parent container


panjenny0:
- [x] progress bar: should be sticky
- [x] parent container: styled div that just needs to have padding to push the form content inwards)

<img width="361" alt="Screen Shot 2020-11-21 at 3 43 13 PM" src="https://user-images.githubusercontent.com/38872354/99890061-45f95880-2c10-11eb-9c31-4dcef85bae38.png">
<img width="729" alt="Screen Shot 2020-11-09 at 10 02 01 AM" src="https://user-images.githubusercontent.com/38872354/98578776-9f59a300-2272-11eb-8047-018d2ce16b2c.png">
https://www.figma.com/file/49ZpQPySx3Rd9hmTvEqOpj/Hacker-Application?node-id=124%3A709

PR #137, opened 2020-11-28
Progress bar

andykwan1124:
References #84. Sticky progress bar centered vertically. Try it out by scrolling all the way down the charcuterie page and click on one of the two buttons.


src/pages/Charcuterie.js
```diff
@@ -53,244 +65,248 @@ export default () => {
...
-        <br></br>
-        <A color="primary" bolded href="https://nwplus.io">
-          This is a hacker app link.
-        </A>
-        <br></br>
-        <br></br>
-        <A color="primary" href="https://nwplus.io">
-          This is an unbolded primary color link.
-        </A>
+      <Container>
```
ianmah, line 72: I don't think we need to create a container here. I'm happy for the progress bar to be inline at the bottom of the charcuterie page

andykwan1124: image

Do you mean something like this?

ianmah: Oh i didn't realize it has absolute positioning. Maybe the component doesnt need positioning? And the consumer can choose how to position it?

panjenny0: Yea I think it's fine if we stick it outside the container for now. After you add the form wrapper in your other PR #138, can you position the sidebar like how it is in the designs and update the usage in this file? or can you somehow make them into one component so it can be easily used?
Screen Shot 2020-11-28 at 6 40 24 PM

andykwan1124: Oh yeah good point. You mean I should remove the positioning in the progress bar component and the consumer would make a wrapping component around it that has the positioning styling right?

panjenny0: wait actually I think we should still have positioning, just because in all of its usecases, it needs to have position set as absolute

panjenny0: scratch what i said about integrating the wrapper and the progress bar - we'll style it later
let's just leave position set on the progress bar

src/pages/Charcuterie.js
```diff
@@ -295,6 +285,14 @@ export default () => {
           throttleTime={1000}
         />
       </FormContainer>
+      <H2>Change Progress Bar</H2>
+      <Button width="flex" onClick={() => progress < 100 && setProgress(progress + 10)}>
+        Increase!
+      </Button>
+      <Button width="flex" onClick={() => progress >= 10 && setProgress(progress - 10)}>
+        Decrease!
+      </Button>
```
panjenny0, line 294: this is kinda cute lol

panjenny0 (approved):

PR #138, opened 2020-11-28
added form padding container

andykwan1124:
References #84. Added form container that only introduces padding. The padding params are assuming the form component ranges from the 'logout' button to the 'back/next' button.

### DIFF
--- a/src/components/Form.js
+++ b/src/components/Form.js
@@ -3,9 +3,9 @@ import styled from 'styled-components'
 import { maxWidthMediaQueries } from './Common'
 
 const Container = styled.div`
-  padding: 1.25em 20em 7.5em 7em;
+  padding: 8vh 16vw;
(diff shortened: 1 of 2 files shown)


ianmah (commented):
Can we use `em` instead of `px`

ianmah:
Why do all of the sides have different paddings?

panjenny0 (changes requested):
Added a suggestion for the wrapper! Also can you add an example of the wrapper in charcuterie? something like this that wraps everything starting fromthe checkboxes
```
<H1>Form Wrapper</H1>
<Form>
...blah blah
</Form>
```

andykwan1124:
> Why do all of the sides have different paddings?
That's mb, I misunderstood what the padding should be applied on lol

panjenny0 (approved):

PR #141, opened 2020-11-29
fix doc.data() firebase bug

acchiang:
prizes were missing call to doc.data(), used reduce instead of filter for data array

image

tfw getting contributions for fixing bugs i introduced myself 😈

### DIFF
--- a/src/containers/Prizes.js
+++ b/src/containers/Prizes.js
@@ -60,8 +60,20 @@ export default () => {
     getPrizes()
       .then(docs => {
         let prizes = {}
-        prizes.main = Object.values(docs.filter(doc => typeof doc.data().place === 'number'))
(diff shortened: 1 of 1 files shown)


src/containers/Prizes.js
```diff
@@ -60,8 +60,20 @@ export default () => {
...
+          docs.reduce((result, doc) => {
+            const data = doc.data()
+            typeof doc.data().place === 'number' && result.push(data)
+            return result
+          }, [])
+        )
+        prizes.sponsor = Object.values(
+          docs.reduce((result, doc) => {
+            const data = doc.data()
+            !doc.data().place && result.push(data)
```
ianmah, line 73: Let's just hope no one puts place = 0

ianmah (approved):

PR #144, opened 2020-11-29
feat: add submission edit form

jackyzha0:
* refactor all form related components into module, can now do `import { Button, TextInput, TextArea, Checkbox, Select, Dropdown } from '../components/Input'`
* pull out getProject logic into `src/utility/firebase.js:getProject(user_id, setProjectCallback, setFeedbackCallback)`
* add basic project edit submission form


src/components/ProjectForm.js
```diff
@@ -0,0 +1,90 @@
...
+}
+
+const FormRow = ({ id, label, children }) => (
+  <div>
+    <Label htmlFor={id}>{label}</Label>
+    <div id={id}>{children}</div>
+  </div>
+)
+
+export default ({ project, submitCallback, onChange }) => {
```
ianmah, line 50: thoughts on using onSubmit

src/pages/Submission/Edit.js
```diff
@@ -1,8 +1,22 @@
...
+import Form from '../../components/ProjectForm'
+
+// TODO: Get from firebase auth
+const USER_ID = 'aIwA36q0kOw7rDDlCkB2'
 
 export default () => {
   const [isSubmissionsOpen, setIsSubmissionsOpen] = useState(false)
+  const [project, setProject] = useState()
+
+  // eslint-disable-next-line
```
ianmah, line 12: can remove?

src/utility/firebase.js
```diff
@@ -37,3 +38,29 @@ export const getSponsors = () => {
       return querySnapshot.docs
     })
 }
+
+export const getProject = async (user_id, setProjectCallback, setFeedbackCallback) => {
```
ianmah, line 42: I was thinking that the Judging page (which displays 5 projects) will need to use something similar to `getProject` but doesn need a feedback callback. Should I just call this function and pass an empty `setFeedbackCallback` or should we split this out to two functions?

src/pages/Submission/Edit.js
```diff
@@ -11,8 +25,17 @@ export default () => {
...
+    console.log(project)
+  }
+
   return (
     // TODO: Implement this page
-    isSubmissionsOpen ? <h1>TODO: Edit submission form</h1> : <h1>Edit - You cannot submit yet</h1>
+    isSubmissionsOpen ? (
+      <Form project={project} submitCallback={submit} onChange={setProject} />
+    ) : (
+      <h1>Edit - You cannot submit yet</h1>
```
ianmah, line 38: idk who's in charge of copy but maybe "Submissions are not open yet. Please check back later." H2?

src/components/ProjectForm.js
```diff
@@ -0,0 +1,90 @@
...
+  return project ? (
+    <div>
+      <H1>Edit "{project.title}"</H1>
+      <TableForm>
+        <FormRow id="title" label="Project Name">
+          <TextInput
+            value={project.title}
+            placeholder="Required"
+            onChange={e => onChange({ ...project, title: e.target.value })}
```
ianmah, line 59: I wonder if passing entire project objects on every keypress could be slow on bad computers. But the code sure is clean... Don't think it needs to be changed but

src/components/Input/index.js
```diff
@@ -0,0 +1,8 @@
+import Button from './Button'
+import Checkbox from './Checkbox'
+import Select from './Select'
+import TextArea from './TextArea'
+import TextInput from './TextInput'
+import ToggleSwitch from './ToggleSwitch'
+import Dropdown from './Dropdown'
+export { Button, Checkbox, Select, TextArea, TextInput, ToggleSwitch, Dropdown }
```
ianmah, line 8: 🧼

src/components/ProjectForm.js
jackyzha0: project object isn't that large so I think it should be fine?

src/utility/firebase.js
jackyzha0: i'm thinking keep the `setFeedbackCallback` empty and we can check for empty callback in the `getProject` function and not execute that part of the code, will keep it cleaner

ianmah (approved):

PR #150, opened 2020-11-30
Add Auth infrastructure

### LINKED ISSUE #108: Login - firebase authentication
Add the authentication logic to #48 
- [x] OAuth for Google
- [x] OAuth for GitHub
- [ ] store user info in db
Screen Shot 2020-11-21 at 2 52 12 PM

Ciuffi:
Closes #108.

This PR adds authentication to the nwPlus portal. 

I would advise reviewing it in this order (or review it commit by commit)

### Auth.js
- It uses react context to handle auth state.
- Login functions take two paremeters, a router callback and a auth state callback (setLocation and setUser)
- Logout function to logout the user

### login.js
- Login functions have just been added to the buttons and the proper auth and router callbacks are passed to them.

### App.js
- A new route type has been added called an AuthRoute. This route will enforce authentication and if it finds the user is not logged in, it will redirect them to the /login page.
- All /application routes have been hidden behind this AuthRoute.

## Using the auth context in your components
- import useAuth from auth.js
- call useAuth in your component E.G.: 
```
const {isAuthed, user, setUser} = useAuth();
```
*Note, most of the time you will only need the user variable*

## user object variables
You can find a full list of variables on the user object here: https://firebase.google.com/docs/reference/js/firebase.User

Relevant ones:
```
user.email //their email
user.displayName //usually their full name (first + last).
user.uid //their user id which we will use to index hacker applications
```

Let me know if anyone needs help getting auth working with their components.

## requesting advice
~currently our login page does nothing if auth fails, should we handle the errors somehow and display it to the user?~
UPDATE: an error popup has been added to fix this


src/App.js
```diff
@@ -63,58 +69,62 @@ function App() {
...
+            </AuthRoute>
+            <AuthRoute path="/application/:part">
+              {params => <ApplicationForm part={params.part} />}
+            </AuthRoute>
+            <PageRoute path="/">
+              <Home />
+            </PageRoute>
+            <AuthRoute path="/application">
+              <Application />
+            </AuthRoute>
```
jackyzha0, line 92: is it possible to use a subrouter for all `/application` routes?

src/utility/Auth.js
```diff
@@ -0,0 +1,55 @@
+import React, { createContext, useContext, useState, useEffect } from 'react'
+import firebase from 'firebase/app'
+import 'firebase/auth'
+
+const AuthContext = createContext()
```
jackyzha0, line 5: why is AuthContext empty here?

src/utility/Auth.js
```diff
@@ -0,0 +1,55 @@
...
+}
+
+export const googleSignIn = async (setUser, setLocation) => {
+  await firebase.auth().setPersistence(firebase.auth.Auth.Persistence.LOCAL)
+  const provider = new firebase.auth.GoogleAuthProvider()
+  try {
+    await firebase.auth().signInWithPopup(provider)
+    const user = firebase.auth().currentUser
+    setUser(user)
+    setLocation('/application')
```
jackyzha0, line 34: seeing as judging will also require users to be signed in, how do we want to handle this redirect? should we pass in a `redirTo` prop?

src/utility/Auth.js
```diff
@@ -0,0 +1,55 @@
...
+  } catch (e) {
+    // Not sure what to do here.
```
jackyzha0, line 36: might be good to multiplex on the error here to provide more info the user, can probably use `TextInputLikeErrorMsg` from `Common.js`

src/App.js
Ciuffi: Yes, but it would really only be for organization (I didn't touch the way its organized because i haven't been part of that planning)

src/utility/Auth.js
Ciuffi: Judging will require more nuanced authentication, so it should probably be a seperate login function

src/utility/Auth.js
Ciuffi: I don't quite understand what you mean, could you go into more detail?

src/utility/Auth.js
Ciuffi: There is no data for the auth context yet. It is initialized in the AuthProvider component.

src/utility/Auth.js
ianmah: It would be nice to display an error, maybe underneath the login buttons that lists the error. I'm trying to login on the preview link and its not working andnot seeing any errors is weird

Ciuffi: I'll add a banner that pops up when the login fails.

Regarding logging in through a preview link: unfortunately its not really possible. Github only allows us to specify one link per app, and firebase only lets us select one app per firebase project. Currently the specified link for the dev firebase is localhost so it can be tested locally, do you think I should change it to amplify?

Ciuffi: (Also, we need to specify the exact url, including subdomains, for github. so it probably won't like the way amplify makes our previews)

src/utility/Auth.js
jackyzha0: what'll be the difference between application auth and registration auth?

src/components/ErorrBanner.js
```diff
@@ -0,0 +1,30 @@
+import React from 'react'
+
+import styled from 'styled-components'
+
+const ErrorDiv = styled.div`
+  max-width: 400px;
+  background-color: ${p => p.theme.colors.error};
+  position: fixed;
+  bottom: ${p => (p.shown ? '10px' : '-100px')};
```
jackyzha0, line 9: is hiding it just shifting it up? should we also set opacity here?

src/pages/Login.js
```diff
@@ -7,6 +7,10 @@ import google from '../assets/icons/google.svg'
 import github from '../assets/icons/github.svg'
 import { useAuth, googleSignIn, githubSignIn } from '../utility/Auth'
 import { useLocation } from 'wouter'
+import { ErrorBanner } from '../components/ErorrBanner'
```
jackyzha0, line 10: typo in name

src/utility/Auth.js
Ciuffi: Oh, i see what you mean. Actually they will be the same, but how will we tell whether someone is going to judging vs application?

src/components/ErorrBanner.js
Ciuffi: I don't think its really necessary since its position is fixed and outside the window.

src/pages/Login.js
Ciuffi: good catch

src/components/ErorrBanner.js
Ciuffi: Actually I'll add it just in case there are some weird styling issues on smaller screens

src/utility/Auth.js
Ciuffi: I've added an error banner.

src/utility/Auth.js
jackyzha0: I was thinking we can default to application. The only time they'll need to auth to judging is if they're already on livesite without being authed and click judging. In this case, we can probably use a parameter like `?redir=judging` or something and parse that to figure out where to redirect them

ianmah:
Awesome PR description 🙏 

If its an easy implementation can we make the error message disappear after a few seconds?

src/utility/Auth.js
Ciuffi: Wouter doesn't have support for search params. I could implement it myself but it sounds a bit like scope creep. Maybe we can have a main landing page after login?

Ciuffi: or choose an already made page to be that landing page

Ciuffi:
> 
> 
> Awesome PR description 🙏
> 
> If its an easy implementation can we make the error message disappear after a few seconds?

not hard at all, I added it.

src/utility/Auth.js
```diff
@@ -0,0 +1,57 @@
...
+}
+
+export const githubSignIn = async (setUser, setLocation) => {
+  await firebase.auth().setPersistence(firebase.auth.Auth.Persistence.LOCAL)
+  const provider = new firebase.auth.GithubAuthProvider()
+  try {
+    await firebase.auth().signInWithPopup(provider)
+    const user = firebase.auth().currentUser
+    setUser(user)
+    setLocation('/application')
```
andykwan1124, line 48: According to Tiffany, we would take the user to the app form if they haven't started or submitted their application yet. So I think we will need to get the user's application status first?

Ciuffi: not a bad idea, we could route them based on application status (if accepted, they'll be routed to judging instead of the application)

andykwan1124: would it be possible to pass the user app status in the AuthContext too? the sidebar display will depend on it

Ciuffi: Yeah its totally possible, I'll add it

Ciuffi: Actually this seems like out of scope for this PR because we also need to create the hacker app document on first sign in. For now, I think this PR has a big enough scope.

src/utility/Auth.js
Ciuffi: hmm looks like we'll do it like this: https://github.com/nwplus/livesite/pull/150#discussion_r532374941

Ciuffi:
### Update regarding routing to different pages/application flow

I will creating another PR to handle application flow, like creating the application in firebase, connecting it to the user and handling user state. Its out of scope for this PR, so please review it with that in mind.

ianmah:
Sounds good! Could you sync w Jenny before you start the second PR, I think someone started working on something similar (if not the same)

panjenny0:
I was going to review this in detail tomorrow but I'll just chime in 😅

@Ciuffi thanks for the update, sounds like a good plan! I think you can go ahead with creating the user document in firebase and etc. in another PR

jackyzha0 (dismissed):
🧞

src/components/ErrorBanner.js
```diff
@@ -0,0 +1,31 @@
...
+  background-color: ${p => p.theme.colors.error};
+  position: fixed;
+  bottom: ${p => (p.shown ? '10px' : '-100px')};
+  opacity: ${p => (p.shown ? '100%' : '0%')};
+  transition: all 0.5s ease-in;
+  margin-left: auto;
+  margin-right: auto;
+  left: 0;
+  right: 0;
+  border: 1px solid grey;
```
panjenny0, line 16: nit: might look better without the border 😝 
```suggestion
```

src/pages/Login.js
```diff
@@ -1,10 +1,16 @@
...
 import Landing from '../containers/Landing'
 import { Button } from '../components/Input'
 import google from '../assets/icons/google.svg'
 import github from '../assets/icons/github.svg'
+import { useAuth, googleSignIn, githubSignIn } from '../utility/Auth'
+import { useLocation } from 'wouter'
+import { ErrorBanner } from '../components/ErrorBanner'
+
+const ErrorMessage =
+  'There was an issue logging you in. If this persists, please contact info@nwplus.io.'
```
panjenny0, line 13: Can we do this, and perhaps put info@nwplus.io into constants and reuse it here?
```suggestion
import { A } from '../components/Typography'

const ErrorMessage = <>
  There was an issue logging you in. If this persists, please contact" 
  <A href="mailto:info@nwplus.io">info@nwplus.io</A>.
</>
```

src/pages/Login.js
```diff
@@ -27,26 +33,57 @@ export const ButtonContainer = styled.div`
...
+  const { setUser } = useAuth()
+  const [, setLocation] = useLocation()
+  const [showError, setShowError] = useState(false)
+
+  const showErrorMessage = error => {
+    if (!error) return
+    setShowError(true)
+    setTimeout(() => {
+      setShowError(false)
+    }, 5000)
```
panjenny0, line 45: Can we increase the timeout here?
```suggestion
    }, 10000)
```

panjenny0 (changes requested):
very nice and it works pretty well! left some tiny suggestions :)

panjenny0 (approved):
looks good boss

src/App.js
```diff
@@ -43,13 +44,21 @@ const PageRoute = ({ path, children }) => {
   )
 }
 
-const NavbarRoute = ({ path, children, name, handleLogout }) => {
+const AuthPageRoute = ({ path, children }) => {
+  const { isAuthed } = useAuth()
+  return <Route path={path}>{isAuthed ? <Page>{children}</Page> : <Redirect to="/login" />}</Route>
+}
+
+const NavbarRoute = ({ path, children }) => {
```
panjenny0, line 52: Quick comment about `NavbarRoute`, since a user's name and the logout button is not displayed on every single screen, can you make it take in `name` and `handleLogout`? 

screens that display name and logout:
- /application/:part
- /application/review

logout only:
- /application/confirmation

no name. no logout:
- /login

ianmah: can we do this in a different PR? I think we should merge this PR Asap

jackyzha0 (approved):

PR #160, opened 2020-12-01
User status and new applicant creation + routing overhaul

Ciuffi:
### User status
This PR adds a user status field to the user object provided by auth. You can access the user status like this
```javascript
user.status // this is the users status
```
The list of available status' are in the Constants.js file under applicantStatus.
### New app creation
When a new user is detected (E.G. no application found for that user) the portal will automatically create a new document with a few fields already set.:
1. The first field set is the email, firstName and lastName fields in BasicInfo
2. The second field set is the _id field which matches the document id as well as the users id.
3. a timestamp will be set for when the document was created

### New redirections
- After logging in, you will be redirected to the correct place. This means `/application` for most statuses, but if we detect that the application status is set to `attending` then you will be redirected to `/judging`.
- Navigating to `/login` while already logged in will now redirect you to the correct place

### Auth loading / routing overhaul
- There was an issue where our authentication would load slightly after our routes did, causing the application to kind of jump around as it receives new information. To fix this, I have added a loading screen for all routes that require authentication and they won't be loaded until auth is finished loading
 **note: I plan to use the same blocking mechanism to also load whether the application is open or not in a later PR**
- Routing has been overhauled because I don't want to block loading of the portal on routes that don't require it. 
   - All Auth routes have been moved to a route at the end of the routing table
   - All Auth routes have been wrapped inside the AuthProvider component
   - Unfortunate it means the 404 component also had to be moved into this route, so it will require a quick load before displaying. (*any suggestions welcome on how to avoid this*)

@panjenny0 You can just stick your loading component inside `auth.js` where the `<>loading..</>` part is

### Testing
- Run this locally (auth doesn't work on previews)
- navigate to either the login page or any of the protected routes (/judging, /application, etc)
- Login, you should be redirected to the `/application` page
- login to firebase and make sure that a new document has been created for you under `/Hackathons/LHD2021/Applicants/{uuid}`
   - Also make sure that your firstName, lastName, email, and _id are set correctly.

Let me know if it works for you and if you have any issues with the routing overhaul.


src/App.js
```diff
@@ -83,72 +92,71 @@ function App() {
...
+            </NavbarAuthRoute>
+            <NavbarAuthRoute path="/application/:part">
+              {params => <ApplicationForm part={params.part} />}
+            </NavbarAuthRoute>
+            <AuthPageRoute path="/judging">
+              <Judging />
+            </AuthPageRoute>
+            <PageRoute path="/judging/admin">
+              <JudgingAdmin />
+            </PageRoute>
```
Ciuffi, line 140: @jackyzha0 this probably needs auth as well, but i'm not sure what kind so i've left it without auth for now.

src/utility/Auth.js
```diff
@@ -10,28 +12,55 @@ export function useAuth() {
...
+  setUser(user)
+  setLocation(getRedirectUrl(status))
+}
+
+export const getRedirectUrl = status => {
+  switch (status) {
+    case applicantStatus.attending:
+      return '/judging'
+    case applicantStatus.applied || applicantStatus.accepted:
+    case applicantStatus.inProgress || applicantStatus.new:
```
acchiang, line 52: nitpick: ik you added these cases because these statuses will likely be used for other routes at some point but i don't think they're needed (and we can add them when it comes to it?)

src/utility/Constants.js
```diff
@@ -15,3 +15,51 @@ export const SOCIAL_LINKS = Object.freeze({
   TW: 'https://www.twitter.com/nwplusubc',
   WEBSITE: 'https://www.nwplus.io',
 })
+
+export const applicantStatus = Object.freeze({
+  attending: 'attending',
+  new: 'new',
+  applied: 'applied',
+  accepted: 'accepted',
+  inProgress: 'inProgress',
```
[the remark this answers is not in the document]
jackyzha0: might be good to do a sync up meeting over this, right now attending/verify only happens with the Discord bot firebase which is separate from the firebase one. would be good to have a proper discussion about what we want to do for login for HC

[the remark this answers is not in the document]
JPGarCar: I can get the bot to change this field from false to true as soon as people mark themselves as attending with the bot. I would just need to add the Tokens and stuff to the bot. Do yall want me to do that?

src/utility/Constants.js
```diff
@@ -15,3 +15,51 @@ export const SOCIAL_LINKS = Object.freeze({
...
+    lastUpdated: '',
+    submitted: false,
+  },
+  status: {
+    applicationStatus: 'inProgress',
+    responded: false, // using responded and attending to check if they un-RSVPed or if they didn't RSVP at all (no RSVP)
+    attending: false, // false for no-rsvp by default
+  },
+  team: '',
+})
```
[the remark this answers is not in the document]
panjenny0: I already had a PR up for the updated statuses but I just quickly added phone numbers too. I think the data types should be the same!
https://github.com/nwplus/database-schemas/pull/22

src/utility/Constants.js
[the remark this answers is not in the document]
Ciuffi: ^I think thats a good idea personally, but lets see what others think.

src/utility/firebase.js
```diff
@@ -92,3 +97,51 @@ export const getProject = async (user_id, setProjectCallback, setFeedbackCallbac
...
+
+  const status = applicant.data().status.applicationStatus
+  if (status === applicantStatus.applied) {
+    return applicantStatus.applied
+  }
+  if (status === applicantStatus.accepted) {
+    return applicantStatus.accepted
+  }
+  return applicantStatus.inProgress
+}
```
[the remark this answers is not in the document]
Ciuffi: If we remove the extra status' then it'll look nicer!

src/utility/Auth.js
Ciuffi: @panjenny0 do we have any plans for different routes for hackers that have already applied or have already been accepted? if not, I'll remove these extra variables.

panjenny0: For hackers who have applied (applied, accepted, rejected, waitlisted), the plan is to redirect them to `/application` -  I think it's ok if you remove this case `case applicantStatus.applied || applicantStatus.accepted:`.
For people who are new or inProgress, it should take them to their first incomplete question (or a new form if they're new) - it would be great if you can add this as well, feel free to do it in a future PR!

src/utility/Constants.js
```diff
@@ -21,3 +21,52 @@ export const SOCIAL_LINKS = Object.freeze({
   TW: 'https://www.twitter.com/nwplusubc',
   WEBSITE: 'https://www.nwplus.io',
 })
+
+export const applicantStatus = Object.freeze({
+  attending: 'attending',
+  new: 'new',
+  applied: 'applied',
+  accepted: 'accepted',
+  inProgress: 'inProgress',
```
panjenny0, line 30: Also on another note, can we add waitlisted, rejected here too? the dashboard (@andykwan1124) has views for waitlist and rejected applicants too

src/utility/firebase.js
panjenny0, line 147: do you think we can just do something like this, using `applicationStatus` from Constants.js?
```
return applicationStatus[status]
```

src/utility/firebase.js
```diff
@@ -92,3 +97,51 @@ export const getProject = async (user_id, setProjectCallback, setFeedbackCallbac
...
+  const newApplication = {
+    ...hackerApplicationTemplate,
+    ...basicInfo,
+    ...submission,
+    ...userId,
+  }
+
+  await applicantsRef.doc(user.uid).set(newApplication)
+}
+
```
panjenny0, line 128: Also was wondering if you can make a simple getter for checking if applicant has responded or not? Just because the dashboard has a view for un-RSVP and no RSVP
```
export const getUserResponse = async user => {
  const applicant = await applicantsRef.doc(user.uid).get()
  if (!applicant.exists) {
      return false;
  }
  return applicant.data().status.responded
}
```

panjenny0 (commented):
Looks solid! Added some comments

src/utility/Auth.js
Ciuffi: Good idea, I think I'll add these changes in a future PR like you suggested!

src/utility/Constants.js
Ciuffi: I think I'll do proper redirecting in a separate PR and update all the status

src/utility/firebase.js
Ciuffi: I don't want to have to support every status because we don't need to, so this function essentially muxes them together

src/utility/firebase.js
Ciuffi: mm I think this should be part of the status, not a separate function, thoughts?

panjenny0: Oh actually sure I think that works too!
I think @andykwan1124 might need these for the dashboard, e.g.
- un-RSVPed: `responded && !attending`
- no RSVP or not attending: `!responded && !attending`

Ciuffi: will include in my status update PR!

panjenny0 (approved):

panjenny0 (approved):

PR #163, opened 2020-12-01
Textarea live char count when max character limit is set

acchiang:
ezgif com-gif-maker (1)

Also fixed a minor bug that showed an undefined maxLength when there is none set in the component props


JPGarCar (dismissed):
Dont know much about this but it looks good!

panjenny0 (approved):
amazing work!!! ✨ also ty for making this so quiccc

ianmah (approved):

PR #166, opened 2020-12-02
Loading

panjenny0:
Loading spinner
- frames per second is too low so it looks not as smooth here
ezgif com-video-to-gif (2)


jackyzha0 (dismissed):

Ciuffi:
@panjenny0 when using this spinner, should I only render the spinner or should I render the spinner + the empty app?

panjenny0:
> @panjenny0 when using this spinner, should I only render the spinner or should I render the spinner + the empty app?

You can just render only the spinner!

Ciuffi (approved):
Very nice, thank you!

PR #168, opened 2020-12-03
add loading spinner to auth

Ciuffi:
Pretty straight forward, loading spinner added to auth while it loads.


Ciuffi:
@panjenny0 got this

PR #169, opened 2020-12-03
Auth for judging admin page

Ciuffi:
### User object changes
- user object now has an admin field that will be true if they are an nwPlus admin or false if they are not.

	```javascript
	user.admin //true if they are an admin
	```
### New auth route
- `<AdminAuthPageRoute />` is a new route that is only accessible by nw admins. 

- Judging admin is now covered by this route.

#### Test 1:
- be not logged in
- try to navigate to /judging/admin
- you should be redirected to /login
#### Test 2:
- be logged in to a NOT nwPlus email
- try to navigate to /judging/admin
- you should be redirected to /application
#### Test 3:
- Be logged into your nwPlus admin email
- try to navigate to /judging/admin
- You should see the admin panel

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -72,6 +72,15 @@ const NoAuthRoute = ({ path, children }) => {
   )
 }
 
+const AdminAuthPageRoute = ({ path, children }) => {
+  const { isAuthed, user } = useAuth()
+  return (
+    <Route path={path}>
+      {isAuthed && user.admin ? <Page>{children}</Page> : <Redirect to="/login" />}
+    </Route>
+  )
+}
+
 function App() {
   useEffect(() => {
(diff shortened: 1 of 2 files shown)


ianmah:
Is the claim added by logging in to cms ?
Also can we make it so the judging admin link in sidebar only shows if user is logged in and admin

Ciuffi:
> 
> 
> Is the claim added by logging in to cms ?
> Also can we make it so the judging admin link in sidebar only shows if user is logged in and admin

Sidebar link seems more related to #170

also yes, you must have logged into the CMS before having access to juding

jackyzha0 (approved):
awesome, this is perfect

PR #171, opened 2020-12-03
Add HackerApplicationContext to handle storing the hacker application + autosaving

Ciuffi:
### what is the hacker application context?
- This is a React Context which allows us to easily access the hacker application from any part of the site where auth is enforced.
- It lets you easily update, save and access the logged in hackers' application. 

### Using the context
Currently I haven't setup the context because i'm waiting for my auth PR to be merged in first. I'll set it up in the routing later. Essentially, anywhere that needs access to the hackerapp must be inside `<AuthProvider>` and `<HackerApplicationProvider>`

To access the hacker app or update it, you should use this hook:
```javascript
const { application, updateApplication, forceSave } = useHackerApplication()
```
- application: the actual application that should be rendered in the app. Holds all the current data for the application.
- updateApplication: should be used like the set function of a useState hook    (e.g.)

   ```javascript
	updateApplication({...application, ...newValues})
	```
- forceSave: This forces the application context to save the current application onto firebase. If this isn't called, the application will be synced every 30 seconds (if changes have been detected)

### Technical details of the context

- The context uses two state variables, one to hold the application and another to hold whether the application has been updated or not.
- When updateApplication is called, it updates the state variable for the application and sets updated to true. 
Every 30 seconds, an interval will check whether the application has been updated. If it has, it will save it to firebase.

### Application opening/closing
- The context will check whether the application is open before displaying it, and if its closed, it'll display a message stating its closed.

**Important note, when the application page first loads, the application might be null. This is because it is still retrieving the application data from firebase. This needs to be handled in the frontend appropriately (loading or otherwise)**

### Testing
The only way to test this currently is to use the react dev tools and check the applicationContext. 
- you should see two boolean state variables. 
- If you set the one that says false to true and wait 30 seconds, it should switch back to false (meaning the autosave worked)
- You should also see your entire application in this context as well.

### DIFF
--- a/src/utility/HackerApplicationContext.js
+++ b/src/utility/HackerApplicationContext.js
@@ -1,0 +1,80 @@
+import React, { createContext, useContext, useState, useEffect, useCallback } from 'react'
+import { useAuth } from './Auth'
+import { getUserApplication, updateUserApplication, getLivesiteDoc } from './firebase'
+import firebase from 'firebase/app'
+import Spinner from '../components/Loading'
+import Page from '../components/Page'
+const HackerApplicationContext = createContext()
+
+export function useHackerApplication() {
+  return useContext(HackerApplicationContext)
+}
+
+export function HackerApplicationProvider({ children }) {
+  const { user } = useAuth()
+  const [application, setApplication] = useState({})
+  const [updated, setUpdated] = useState(false)
+  const [applicationOpen, setApplicationOpen] = useState(null)
+
+  /**Initialize retrieval of hacker application */
+  useEffect(() => {
+    const retrieveApplication = async () => {
+      const app = await getUserApplication(user.uuid)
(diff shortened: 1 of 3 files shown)


src/utility/HackerApplicationContext.js
```diff
@@ -0,0 +1,86 @@
...
+      ...application,
+      submission: {
+        lastUpdated: firebase.firestore.Timestamp.now(),
+        submitted: false,
+      },
+    }
+    await updateUserApplication(user.uid, updatedApp)
+    setApplication(updatedApp)
+    setUpdated(false)
+  }, [application, user])
```
acchiang, line 43: woohoo @anneguo3 check this out

jackyzha0 (approved):
super clean, love the comments 🧼

PR #176, opened 2020-12-04
Analytics

Ciuffi:
### What do Analytics do

Analytics allow us to track a user as they use our website. We can define certain events and trigger them through javascript on our sites. These events are then aggregated with the users ID and gives us valuable insights on what worked for that user and what maybe didn't work for that user.

### How to use analytics
Simply import the analytics package from our firebase file and use the logEvent command.

```javascript
import { analytics } from '../utility/firebase'
analytics.logEvent(<eventName>, <payload>)
// Example
analytics.logEvent(ANALYTICS_EVENTS.Logout, { userId: user.uid })
```
### Current events
Here is a list of the events I have added and what they do:
- **Login**: Called when the user logs in.
- **AccessApplication**: Called when the user accesses their application.
- **EditApplication**: Called when the editApplication button is pressed.
- **Signup**: Called when a user first signs up and their application is made.
- **SocialMediaConversion**: Called when someone clicks on a social media link.
	- payload includes which social media they're navigating to.
- **Logout**: Called when the user logs out.
- **NotificationToggled**: Called when the user toggles notifications.
	- payload: enabled/disabled

### Important note before merging
We might want to wait until after work on the livesite slows down. This change when merged will cause most peoples dev envs to crash because our env file doesn't include an appId or a measurementId.


panjenny0 (approved):
Looks good! Approving now but let's merge later to prevent breaking people's envs

acchiang (approved):
rip merge conflicts again heh

andykwan1124 (commented):
woww firebase analytics is really cool!!!! I tested the events and can confirm that `SocialMediaConversion`, `Login`, and `AccessApplication` work. But I don't see the other events on firebase, do u know where I can find them?

panjenny0 (approved):

PR #181, opened 2020-12-04
change login copy to Return to Portal

acchiang:
literally just a text change: login page includes link that says Return to Portal if user doesnt want to login 
ezgif com-gif-maker (2)

### DIFF
--- a/src/pages/Login.js
+++ b/src/pages/Login.js
@@ -89,7 +89,7 @@ export default () => {
             Continue with GitHub
           </StyledButton>
         </ButtonContainer>
-        <A href="/">Skip For Now</A>
+        <A href="/">Return to Portal</A>
       </Landing>
       <ErrorBanner shown={showError} message={ErrorMessage} />
     </>


src/pages/Login.js
```diff
@@ -89,7 +89,7 @@ export default () => {
             Continue with GitHub
           </StyledButton>
         </ButtonContainer>
-        <A href="/">Skip For Now</A>
+        <A href="/">Return to Portal</A>
```
jackyzha0, line 92: can we do a ternary on `copyText.hackathonName` for this?

jackyzha0: isn't the Skip for Now for nwHacks?

Ciuffi: We just decided to change the copy, both will be the same

acchiang: no there is only one login.js page that changes appearance based on them, so both HC and nw have this link rn 

this gives all users an option to go back to dashboard if they decide they dont' want to proceed 

if you want to remove this link here (I admit I was a bit presumptuous when i made it) i'm cool with it, just let me know if Logout should redirect to Login page or to home page

Ciuffi: + ternary probably too complicated once we run 4 hackathons..

Ciuffi (approved):
Could you make your PRs smaller? I don't like reviewing such huge file diffs

acchiang:
> Could you make your PRs smaller? I don't like reviewing such huge file diffs

maybe you can include this as a retro item during your next sprint meeting
oh wait-

PR #187, opened 2020-12-05
Rework hacker application context

Ciuffi:
### Main changes
After doing some more research on `useCallback` and `useEffect`, I noticed that my useEffects were being called more often than they should. This is because the application was a dependency and it would cause my useEffects to run every time the application is updated. 

To fix this, I no longer access `application` in `forceSave`, removing it from the dependency tree. Instead I rely on a reference to the application in order to save it. This reference is updated every time the application is updated.

I also removed the dependency on `update`, using `setUpdate` to access it dynamically instead.

### Application merging
Applications are now merged properly so that the burden doesn't fall on the person creating the application. You may now use `updateApplication` like this:
```javascript
updateApplication({
	basicInfo,
	status,
	etc
})
```
merging will be handled for you.
@panjenny0 Please make sure to let people working on the application know!

### Save on navigate away
Because the hackerContext no longer encompasses all application pages, the context will now attempt to save your application as soon as its dismounted. This means if you're on your application and click on any link to another page, it should save the application for you.

**note this will cause errors because state updates are being called from `forceSave`, even though the context is dismounted. This is fine. It's not actually a memory leak and `forceSave` is cleaned up properly after saving.**

**Another note, this will not save the application if the page is closed. Only if a wouter navigation link is clicked.**

### DIFF
--- a/src/utility/HackerApplicationContext.js
+++ b/src/utility/HackerApplicationContext.js
@@ -1,4 +1,4 @@
-import React, { createContext, useContext, useState, useEffect, useCallback } from 'react'
+import React, { createContext, useContext, useState, useEffect, useCallback, useRef } from 'react'
 import { useAuth } from './Auth'
 import { getUserApplication, updateUserApplication, getLivesiteDoc } from './firebase'
 import firebase from 'firebase/app'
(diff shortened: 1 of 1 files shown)


jackyzha0 (approved):
love the comments, lgtm 🧼

PR #192, opened 2020-12-05
admins view grades

ianmah:
image

### DIFF
--- a/src/containers/JudgingPanel.js
+++ b/src/containers/JudgingPanel.js
@@ -1,9 +1,9 @@
-import React, { useRef, useState } from 'react'
+import React, { useRef, useEffect, useState } from 'react'
 import { Button } from '../components/Input'
 import { H1, H3, P, A } from '../components/Typography'
 import { Card } from '../components/Common'
 import Accordion from '../components/Accordion'
-import { syncToFirebase } from '../utility/firebase'
+import { syncToFirebase, projectsRef } from '../utility/firebase'
 import ProjectTable from '../components/Judging/ProjectTable'
 import SponsorSubmissions from '../components/Judging/SponsorSubmissions'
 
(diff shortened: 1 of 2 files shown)


src/containers/JudgingPanel.js
```diff
@@ -46,16 +47,67 @@ class Project {
   }
 }
 
+const getGradedProjects = async () => {
+  const projectDocs = await projectsRef.get()
+  const projectData = projectDocs.docs.map(projectDoc => {
+    const project = projectDoc.data()
+    if (project.grades) {
+      project.countGraded = Object.values(project.grades).length
```
jackyzha0, line 55: thoughts on doing `Object.values()` in `formatProject`? I'm doing this cast too :(

src/containers/JudgingPanel.js
```diff
@@ -46,16 +47,67 @@ class Project {
...
+          project[key] = project[key] ? project[key] + value : value
+          project.total = project.total ? project.total + value : value
+        })
+      })
+      project.grade = project.total / project.countGraded / 5
+      project.tech = project.tech / project.countGraded
+      project.design = project.design / project.countGraded
+      project.functionality = project.functionality / project.countGraded
+      project.creativity = project.creativity / project.countGraded
+      project.pitch = project.pitch / project.countGraded
```
jackyzha0, line 68: can probably be rewritten to use a reduce which would be a lot cleaner (ok for now cause we're time crunched, but maybe add a TODO and make a ticket)

src/containers/JudgingPanel.js
```diff
@@ -46,16 +47,67 @@ class Project {
...
+    return project
+  })
+  projectData.sort((a, b) => {
+    if (a.grade > b.grade) {
+      return -1
+    }
+    if (a.grade < b.grade) {
+      return 1
+    }
+    return 0
```
jackyzha0, line 82: pretty sure you can just return `a.grade - b.grade`

acchiang: @jackyzha0 i think u mean `b.grade - a.grade` according to the logic here 
but ya @ianmah `.sort()` only cares about positive vs negative, doesn't have to be -1/0/1

src/containers/JudgingPanel.js
ianmah: how do write a reduce for average?

src/containers/JudgingPanel.js
ianmah: this pr is giving me major 310 flashbacks

src/containers/JudgingPanel.js
ianmah: like in general or here? I am not using `formatProject` here, should I be?

src/containers/JudgingPanel.js
acchiang: lmao truth 
tbh these are minor comments, i think it's good to merge (up to u if u wanna make refactor issues so we dont' forget -- or so we can close the refactor issues later if we decide not to do it)

acchiang (dismissed):
yuh (crunch time!)

src/containers/JudgingPanel.js
ianmah: note: am not iterating thru project because project is populated with all kinds of non grading related data such as devpostUrl, description etc

Could be refactored so it doesn't but i'll save that for later

src/containers/JudgingPanel.js
```diff
@@ -72,15 +72,7 @@ const getGradedProjects = async () => {
...
-  projectData.sort((a, b) => {
-    if (a.grade > b.grade) {
-      return -1
-    }
-    if (a.grade < b.grade) {
-      return 1
-    }
-    return 0
-  })
+  projectData.sort((a, b) => b.grade - a.grade)
```
acchiang, line 75: love that for us

acchiang (approved):

jackyzha0 (approved):
🧼

PR #202, opened 2020-12-06
TEMPORARILY fix IOS mobile Notifications crash bug 

dryu99:
Disabled following uses of Notifications API when livesite is accessed on an IOS mobile device:
- alerting user with new announcements through push notifications
- notification toggle initialization logic

Result:
- Behaviour unchanged on desktop site + android devices? (don't own an android device so didn't test, plz enlighten me if there is reliable way to do so on desktop) 
- Notification toggle no longer visible on IOS mobile devices 
- no more crashing on mobile IOS devices 🤞 

Notes:
Correct me if I'm wrong, but it doesn't look like mobile IOS supports push notifications at all. Not sure what a proper fix would look like atm.

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -24,7 +24,7 @@ import {
 } from './pages'
 import Page from './components/Page'
 import { db } from './utility/firebase'
-import { DB_COLLECTION, DB_HACKATHON } from './utility/Constants'
+import { DB_COLLECTION, DB_HACKATHON, IS_DEVICE_IOS } from './utility/Constants'
 import notifications from './utility/notifications'
 import { AuthProvider, getRedirectUrl, useAuth } from './utility/Auth'
 import { HackerApplicationProvider } from './utility/HackerApplicationContext'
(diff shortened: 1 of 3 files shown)


src/App.js
```diff
@@ -110,22 +110,27 @@ const JudgingViewContainer = ({ params }) => {
 }
 
 function App() {
+  // TODO create reusable Announcements firebase ref in firebase.js to avoid redundant fb calls in Announcements.js
```
jackyzha0, line 113: noted

jackyzha0 (approved):

src/App.js
dryu99: was my bad 😅

src/App.js
```diff
@@ -110,22 +110,27 @@ const JudgingViewContainer = ({ params }) => {
...
   useEffect(() => {
-    const unsubscribe = db
-      .collection(DB_COLLECTION)
-      .doc(DB_HACKATHON)
-      .collection('Announcements')
-      .orderBy('timestamp', 'desc')
-      .onSnapshot(querySnapshot => {
-        // firebase doc that triggered db change event
-        const changedDoc = querySnapshot.docChanges()[0]
+    // don't notify users on IOS devices because Notification API incompatible
```
Ciuffi, line 115: so lame

PR #204, opened 2020-12-06
Added terms and conditions to HackerApplicationContext

panjenny0:
After working on the review page, I realized that we need a way to keep track of whether or not the applicant has accepted the terms and conditions at the end of the application.
As such, I've updated the application context with some new fields:
```
termsAndConditions: {
    MLHCodeOfConduct: false,  // do they accept the MLH code of conduct
    MLHPrivacyPolicy: false,    // do they accept the MLH privacy policy
    shareWithnwPlus: false,   // do they allow their data to be shared with us
    shareWithSponsors: false,   // do they allow us to share their info with sponsors
  },
```
p.s. @Ciuffi I hope this doesn't break anything!
and FYI @vincentchiang1 this probably means that autosave should be included in the review page as well 😬 
Screen Shot 2020-12-05 at 10 49 00 PM

### DIFF
--- a/src/utility/Constants.js
+++ b/src/utility/Constants.js
@@ -78,5 +78,11 @@ export const hackerApplicationTemplate = Object.freeze({
     responded: false, // using responded and attending to check if they un-RSVPed or if they didn't RSVP at all (no RSVP)
     attending: false, // false for no-rsvp by default
   },
+  termsAndConditions: {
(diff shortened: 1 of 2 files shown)


vincentchiang1 (approved):

Ciuffi (approved):
This should work fine, LGTM!

PR #214, opened 2020-12-09
App part 2

anneguo3:
Added pulling of data from db and ability to upload resumes to firebase storage
Resumes for each applicant are stored under their unique ID, thus if they upload a new resume it will rewrite over the existing document rather than us storing numerous docs per person

Note: found weird error in part-1 from the mapping of formInputs.ethnicity @Ciuffi i think you added those changes?
Screen Shot 2020-12-09 at 3 37 31 PM

TODO: add in validation for part 2 fields


panjenny0:
> Note: found weird error in part-1 from the mapping of formInputs.ethnicity @Ciuffi i think you added those changes?
> Screen Shot 2020-12-09 at 3 37 31 PM

Oh I added the map, let me look into this! I'll open a PR with a fix.

panjenny0 (approved):
Looks good! 🚀

PR #232, opened 2020-12-15
Handle ignorable firebase auth errors

panjenny0:
### Ignoring certain errors
Added error-handling of `cancelled-popup-request` as observed by @dryu99 
Also added `user-cancelled` error to the ignore list, since this may get thrown as well when the user closes the popup

image

### Slight change to `ErrorDiv`
Screen Shot 2020-12-14 at 9 02 29 PM

### DIFF
--- a/src/pages/Login.js
+++ b/src/pages/Login.js
@@ -6,7 +6,7 @@ import { Button } from '../components/Input'
 import google from '../assets/icons/google.svg'
 import github from '../assets/icons/github.svg'
 import { useAuth, googleSignIn, githubSignIn } from '../utility/Auth'
-import { DB_HACKATHON } from '../utility/Constants'
+import { DB_HACKATHON, firebaseAuthError } from '../utility/Constants'
 import { useLocation } from 'wouter'
 import ErrorBanner from '../components/ErrorBanner'
 import { A } from '../components/Typography'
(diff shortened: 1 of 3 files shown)


src/pages/Login.js
```diff
@@ -86,7 +101,7 @@ export default () => {
         </ButtonContainer>
         {DB_HACKATHON === 'LHD2021' && <A href="/">Return to Portal</A>}
       </Landing>
-      <ErrorBanner>{error ? <ErrorMessage message={error.message} /> : null}</ErrorBanner>
+      <ErrorBanner>{error ? handleAuthError(error.code, error.message) : null}</ErrorBanner>
```
kevin-zou, line 104: Not sure what the possible states of `error` are, but would
```
{error && <ErrorBanner>{handleAuthError(erro.code, error.message)}</ErrorBanner>
```
be cleaner? Feels weird setting it to `null` otherwise and then setting the visibility of it based on the lack of children. Fine with this though!

kevin-zou (approved):
Nice! Let a quick comment but nothing required to release

kozr:
What was the problem with logging in with a different credential provider?

src/pages/Login.js
panjenny0: Ohh for certain errors, we don't want to show the error banner at all, so that's why I made `handleAuthError` decide if we should show the banner

kozr (approved):

panjenny0:
> What was the problem with logging in with a different credential provider?

I don't know tbh, my guess is on the firebase auth side it's not ideal/may cause unexpected issues?

PR #234, opened 2020-12-15
Some changes and fixes to resume uploading

Ciuffi:
### Main changes
- The upload happens as soon as the user selects the resume.
- The resume name is now set after the upload completes.
- `handleResume` now handles validating the resume name and the file size before uploading.
- resume name is now taken from `file.name` rather than `e.target.value`, giving us names without weird prefixes. (e.g. `my-resume.pdf` vs `c:\fakedir\my-resume.pdf`)
- I added a message in the `ResumeUploadButton` hint that states the resume was successfully uploaded.

   image

   **(Note, if an upload fails and a resume was previous uploaded, the name will stay and an error will also be shown. This is intended as the original resume is still uploaded.)**
### Fixes
- The review page was crashing because the regex string used to remove the prefix from the file name didn't like a `#` symbol in the file name.
- Now that we no longer have the file prefix, we no longer have the need for the match string, so I removed it.

This should fix the issues with the review page crashing as well as handle the file upload in a cleaner way.

### How to test
1. navigate to the preview URL.
2. Start filling out the application form.
3. On part two, try to upload these cases:
   A) A file over 2mb.
   B) A file under 2mb but not a pdf.
   C) A file that is a PDF under 2mb.
4. Make sure that you get errors if you try option A and option B.
	- The errors should be the same.
	- the name of the file you are uploading should not show up next to the upload button.
5. make sure that you get no errors when you try option C
6. make sure that the name of the file you uploaded when trying option C is shown next to the upload button (as soon in the screenshot above)
7. make sure that the name of the file you uploaded in option C shows up in the review oprtion of the application under the skills section.

### DIFF
--- a/src/components/ApplicationForm/ReviewCards.js
+++ b/src/components/ApplicationForm/ReviewCards.js
@@ -197,14 +197,7 @@ export default ({ formInputs, handleEdit, onChange, errors }) => {
         </JohnDiv>
         <StyledBanner wide={true} blur>
           <ContentWrapper>
-            <InfoGroup
-              heading="Resume"
-              data={
-                formInputs.skills.resume
-                  ? formInputs.skills.resume.match(/[/\\]([\w\d\s.\-()]+)$/)[1]
-                  : ''
-              }
-            />
+            <InfoGroup heading="Resume" data={formInputs.skills.resume ?? ''} />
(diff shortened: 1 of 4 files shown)


panjenny0 (approved):
Looks good thank you!!

acchiang:
very good testing instructions themk

PR #244, opened 2020-12-20
fix sponsor tier name check bug

dryu99:
Fixed bug where inkind sponsor logos were appearing on sponsors logo page (this shouldn't be happening). Bug was occurring because of tier name inconsistencies. We are now casting tier names to lowercase before running through relevant logic (e.g. Inkind vs inkind). 

Also moved `getSponsors` from firebase utils file to sponsors logo component to improve cohesion (I should've done this before 🙉 )

### DIFF
--- a/src/containers/SponsorLogos.js
+++ b/src/containers/SponsorLogos.js
@@ -1,12 +1,24 @@
 import React, { useEffect, useState } from 'react'
 import styled from 'styled-components'
-import { getSponsors } from '../utility/firebase'
 import Sponsors from '../components/SponsorLogos'
+import { db } from '../utility/firebase'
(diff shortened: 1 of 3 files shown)


src/containers/SponsorLogos.js
```diff
@@ -1,20 +1,34 @@
...
+const getSponsors = () => {
+  return db
+    .collection(DB_COLLECTION)
+    .doc(DB_HACKATHON)
+    .collection('Sponsors')
+    .get()
+    .then(querySnapshot => {
+      return querySnapshot.docs
+    })
+}
```
jackyzha0, line 21: thank you for moving this to the container 🙏

jackyzha0 (approved):

PR #248, opened 2020-12-30
Use rsvp flag

### LINKED ISSUE #240: Disable RSVP after RSVP deadline
- [ ] add canRSVP in cms (similar to applicationOpen feature flag)
- [ ] disable RSVP after deadline

andykwan1124:
This PR closes #240.

## What changed

1. The RSVP button's state now depends on the `rsvpOpenFlag` in the CMS. Devs can now use the CMS to enable/disable RSVP in the feature flag section of nwHacks2021.
2. If the `rsvpOpenFlag` is toggled off and the applicant was accepted but still has not responded, the sidebar and dashboard will display the application status `NO RSVP`.
3. Some nit changes

## How to test

For the first change:
1. Make sure your entry in the `Applicant` collection has a status of `applied`  
2. Go to the CMS and toggle the `rsvpOpenFlag`, you should see the RSVP button being enabled/disabled when the flag is active/inactive

For the second change: 
1. Make sure your entry has a status of `applied`, `responded: false` and `attending: false`
2. Toggle the `rsvpOpenFlag`, you should see `NO RSVP` on the dashboard and sidebar

For the third change (general case):
Try out all the possible applicant states just in case

### DIFF
--- a/src/components/ApplicationDashboard.js
+++ b/src/components/ApplicationDashboard.js
@@ -209,6 +209,8 @@ const Dashboard = ({
   username,
   editApplication,
   relevantDates,
+  isRsvpOpen,
+  isLoadingAppStatus,
 }) => {
   return (
     <Container>
@@ -230,7 +232,14 @@ const Dashboard = ({
       <StatusContainer>
         <div>
           <AppStatusText>
-            Application status: {hackerStatuses(relevantDates, username)[hackerStatus]?.cardText}
+            Application status:{' '}
+            {hackerStatus === 'acceptedNoResponseYet'
+              ? !isLoadingAppStatus
+                ? isRsvpOpen
+                  ? hackerStatuses()[hackerStatus]?.cardText
+                  : 'NO RSVP'
+                : null
+              : hackerStatuses()[hackerStatus]?.cardText}
           </AppStatusText>
           <StatusBlurbText>
             {hackerStatuses(relevantDates, username)[hackerStatus]?.blurb}
@@ -240,10 +249,10 @@ const Dashboard = ({
           <SocialMediaLinks />
           <RSVPButton
             width="flex"
-            onClick={isApplicationOpen && (() => setRSVP(canRSVP))}
+            onClick={isRsvpOpen && (() => setRSVP(canRSVP))}
             shouldDisplay={canRSVP || hackerStatus === 'acceptedAndAttending'}
             color={canRSVP ? 'primary' : 'secondary'}
-            disabled={!isApplicationOpen}
+            disabled={!isRsvpOpen}
           >
             {canRSVP ? 'RSVP' : 'un-RSVP'}
           </RSVPButton>
--- a/src/components/ApplicationDashboard.js
+++ b/src/components/ApplicationDashboard.js
@@ -156,9 +156,24 @@ export const hackerStatuses = (relevantDates, hackerName = null) => ({
     ),
     blurb: `We can't wait to see you at ${copyText.hackathonName}! You'll be receiving another email closer to the event date with more information regarding the schedule and other logistics. If you find out you can't make it to ${copyText.hackathonName} anymore due to a change in your schedule, please update your RSVP status so we can allocate spots for waitlisted hackers!`,
   },
-  acceptedNotAttending: {
+  acceptedUnRSVP: {
     sidebarText: "Un-RSVP'd",
     cardText: "Un-RSVP'd",
+    blurb: (
+      <>
+        We're sorry you won't be attending {copyText.hackathonName}. We do hope to see you at our
+        future events, visit our site{' '}
+        <A bolded color="primary" href={SOCIAL_LINKS.WEBSITE}>
+          nwplus.io
+        </A>{' '}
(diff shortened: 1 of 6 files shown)


src/components/ApplicationDashboard.js
```diff
@@ -230,7 +232,14 @@ const Dashboard = ({
...
           <AppStatusText>
-            Application status: {hackerStatuses(relevantDates, username)[hackerStatus]?.cardText}
+            Application status:{' '}
+            {hackerStatus === 'acceptedNoResponseYet'
+              ? !isLoadingAppStatus
+                ? isRsvpOpen
+                  ? hackerStatuses()[hackerStatus]?.cardText
+                  : 'NO RSVP'
+                : null
+              : hackerStatuses()[hackerStatus]?.cardText}
```
panjenny0, line 242: Screen Shot 2020-12-30 at 4 43 28 PM
Screen Shot 2020-12-30 at 4 47 37 PM

The message for No RSVP should be the same as that of un-RSVP, except instead of `un-RSVP'd`, it should say `No RSVP` - can you make this change? 
 
Also for code quality and better separation of concerns, it would be better if this logic was handled in `hackerStatuses` - i.e. in the parent container, have logic to set the status to `acceptedNoRSVP`, then add another case for `acceptedNoRSVP` in `hackerStatuses` and rename `acceptedNotAttending`.

```suggestion
            Application status: {hackerStatuses()[hackerStatus]?.cardText}
```

src/components/Sidebar.js
```diff
@@ -141,11 +143,18 @@ export default ({
...
-              <StatusText>{hackerStatuses()[hackerStatus]?.sidebarText}</StatusText>
+              <StatusText>
+                {hackerStatus === 'acceptedNoResponseYet'
+                  ? !isLoadingAppStatus
+                    ? isRsvpOpen
+                      ? hackerStatuses()[hackerStatus]?.sidebarText
+                      : 'NO RSVP'
+                    : null
+                  : hackerStatuses()[hackerStatus]?.sidebarText}
+              </StatusText>
```
panjenny0, line 157: This logic seems to be duplicating that of `components/ApplicationDashboard`. As well, `Sidebar` should only handle presentation logic of the UI. I recommend moving this logic to the parent container and adding an additional case handling no RSVP in the `hackerStatuses`

src/containers/Application/Dashboard.js
```diff
@@ -63,15 +70,21 @@ const ApplicationDashboardContainer = () => {
   }, [setLivesiteDoc])
```
panjenny0, line 71: instead of handling loading in the child components, I think you can do something like this and remove the loading prop that is passed down. 
```
  if (isLoadingAppStatus) return <Spinner />
```

src/containers/Application/Dashboard.js
```diff
@@ -31,27 +40,25 @@ const ApplicationDashboardContainer = () => {
...
     hackerStatusObject !== undefined &&
     (hackerStatusObject.applicationStatus === 'accepted'
       ? hackerStatusObject.responded
         ? hackerStatusObject.attending
           ? 'acceptedAndAttending'
           : 'acceptedNotAttending'
         : 'acceptedNoResponseYet'
+      : hackerStatusObject.applicationStatus === 'scored'
+      ? 'applied'
       : hackerStatusObject.applicationStatus)
```
panjenny0, line 53: Nit for code quality: Can you see if you can convert this into if conditions instead of using nested ternary operators? Having too many nested ternary conditions tends to make the code not as maintainable.

panjenny0 (changes requested):
Thanks for adding this Andy! Main thing: No RSVP should display the same message as un-RSVP'd. Also left some comments for code quality.

src/containers/Application/Dashboard.js
```diff
@@ -30,28 +40,33 @@ const ApplicationDashboardContainer = () => {
...
+  } else if (applicationStatus === 'scored') {
+    hackerStatus = 'applied'
+  } else {
+    hackerStatus = applicationStatus
+  }
 
-  const canRSVP =
-    hackerStatus === 'acceptedNoResponseYet' || hackerStatus === 'acceptedNotAttending'
-  const setRSVP = canRSVP => {
+  const canRSVP = hackerStatus === 'acceptedNoResponseYet' || hackerStatus === 'acceptedUnRSVP'
```
panjenny0, line 64: It seems like the RSVP button is not displayed on the No RSVP dashboard - even though they technically missed their chance to RSVP, based on the designs, I think we should still display a disabled RSVP button. Can you modify this? 
```suggestion
  const canRSVP = hackerStatus === 'acceptedNoResponseYet' || hackerStatus === 'acceptedUnRSVP' || hackerStatus === 'acceptedNoRSVP'
```
Screen Shot 2020-12-31 at 5 18 29 PM
Screen Shot 2020-12-31 at 5 19 33 PM

panjenny0 (changes requested):
One small thing then I think this is good to merge! Good job refactoring your changes :)

panjenny0 (approved):

PR #258, opened 2021-01-01
Fix submission errors

Ciuffi:
This should fix the submission issues we've been seeings.

### What changed
- The application waits to be saved before it continues on to the thanks for applying page.
- A loading indicator was added next to the submit/next page buttons to show that the form is being saved.

### How to test
- Just make sure that the buttons still work
- The indicator should only show up when moving to the next page or uploading a resume 
- the app should always submit

Hopefully this fixes all the issues


panjenny0 (approved):
lgtm! thanks for fixing this 🏅

PR #260, opened 2021-01-02
Only set applicationStatus to applied if current status is inProgress

panjenny0:
## What was changed
- changed submit function to only set `applicationStatus` to `applied` if the current status is `inProgress`

#### making sure `inProgress` status is set to `applied` upon submission
- go to `/application/review` on portal
- set your application status to `inProgress` in the db
- submit your application
- check the db - status should update to `applied` and dashboard should say awaiting assessment
#### making sure statuses other than `inProgress` is not set to applied after submission
- set your status in the db to `accepted`
- go to `application/review` to submit again
- status should not change

### DIFF
--- a/src/containers/Application/Review.js
+++ b/src/containers/Application/Review.js
@@ -41,14 +41,13 @@ export default () => {
       window.alert('Please agree to the required terms and conditions.')
       return
     }
-    updateApplication({
-      status: {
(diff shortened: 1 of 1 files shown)


andykwan1124 (approved):

Ciuffi:
why is this change being made?

PR #264, opened 2021-01-07
Remove grade

### LINKED ISSUE #253: Admins remove a grade
Add button in table from #247 for admins to remove a grade

ianmah:
closes #253

### DIFF
--- a/src/containers/JudgingPanel.js
+++ b/src/containers/JudgingPanel.js
@@ -9,7 +9,6 @@ import { ProjectTable, ProjectGradeTable, GradeTable } from '../components/Judgi
 import SponsorSubmissions from '../components/Judging/Admin/SponsorSubmissions'
 import ProgressBar from '../components/ProgressBar'
 import { JUDGING_RUBRIC, calculateGrade } from '../utility/Constants'
-import { useAuth } from '../utility/Auth'
 
 class CSV {
   constructor(data) {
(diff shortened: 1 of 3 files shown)


ianmah:
If there is no "user" aka "Submitted by" field on the grade it will error. We can assume that that will never happen on prod?

jackyzha0:
> If there is no "user" aka "Submitted by" field on the grade it will error. We can assume that that will never happen on prod?

we can probably just fain silently and log the error

src/containers/JudgingPanel.js
```diff
@@ -108,24 +108,30 @@ const calculateResiduals = project => {
 
 const getProjectData = async () => {
   const projectDocs = await projectsRef.get()
-  return projectDocs.docs.map(projectDoc => projectDoc.data())
+  return projectDocs.docs.map(projectDoc => {
+    return { ...projectDoc.data(), id: projectDoc.id }
```
jackyzha0, line 112: why is this explicit spread needed?

src/containers/JudgingPanel.js
```diff
@@ -108,24 +108,30 @@ const calculateResiduals = project => {
...
+  })
 }
 
 const getGrades = async () => {
-  const gradeData = (await getProjectData()).reduce((acc, project) => {
+  const gradeData = []
+  const projectData = await getProjectData()
+  projectData.forEach(project => {
```
jackyzha0, line 119: would filter make more sense here

jackyzha0 (approved):
some small comments, everything else looks good chief 🎥

src/containers/JudgingPanel.js
ianmah: not sure because filter takes a predicate so id have to do a filter then a foreach?

src/containers/JudgingPanel.js
ianmah: data doesnt come with id and id doesnt come with data

jackyzha0 (approved):
concerns addressed over text 💱

PR #265, opened 2021-01-08
Implement disqualify button

### LINKED ISSUE #255: Ability to disqualify a team
Add a button into the main project table to allow admins to manually disqualify teams. This will be an internal soft-ban and will show their submission row with a red background.

Add the ability to filter the table to top `n` entries and hide disqualified entries.

ianmah:
closes #255

### DIFF
--- a/src/containers/JudgingPanel.js
+++ b/src/containers/JudgingPanel.js
@@ -1,6 +1,6 @@
 import React, { useRef, useEffect, useState } from 'react'
 import { MoonLoader } from 'react-spinners'
-import { db, syncToFirebase, projectsRef, submitGrade } from '../utility/firebase'
+import { syncToFirebase, projectsRef, submitGrade } from '../utility/firebase'
 import { Button, ToggleSwitch } from '../components/Input'
 import { H1, H3, P, A } from '../components/Typography'
 import { Card } from '../components/Common'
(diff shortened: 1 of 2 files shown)


src/containers/JudgingPanel.js
```diff
@@ -202,6 +202,23 @@ export default () => {
...
+          alert('Project does not exist')
+          return
+        }
+        const disqualified = !projectDoc.data().disqualified
+        transaction.update(projectsRef.doc(projectId), { disqualified })
+      })
+    } catch (e) {
+      alert(e)
+    }
+    window.location.reload()
```
jackyzha0, line 219: is there no way to refresh the data without reloading?

src/containers/JudgingPanel.js
```diff
@@ -202,6 +202,23 @@ export default () => {
     await submitGrade(id, { ...score, removed: true }, { uid: gradeId, email: score.user })
   }
 
+  const onDisqualify = async projectId => {
+    try {
+      await db.runTransaction(async transaction => {
```
jackyzha0, line 207: if this is a single update operation, why do we need a transaction here?

jackyzha0 (dismissed):
small questions but overall looks good chief 🧟‍♂️

src/containers/JudgingPanel.js
ianmah: i was lazy cuz its admin - whats the easiest way to refresh you think?

jackyzha0: honestly this is fine LOL iTs aDmiN

src/containers/JudgingPanel.js
ianmah: Oh I thought we would need a transaction since we are reading and then writing and there could be race conditions. But maybe its better if we just do what the UI shows instead of what the most up to date field is in the backend? (admin isn't subscribed to judging info)

jackyzha0: i think keeping it consistent with UI makes sense here

ianmah: agreee

jackyzha0 (approved):
very pog 🐿

PR #282, opened 2021-11-02
Configure for HackCamp 2021

kevin-zou:
## Description
Configuring theme for HackCamp 2021:
- background color
- secondary color
- schedule event tag colors
- collection names

Let me know if there's anything else I need to configure to get this ready for HackCamp!


anneguo3 (approved):
LGTM! 🎉

probably worth getting Ian or Jacky's review on it

public/index.html
```diff
@@ -20,7 +20,7 @@
       work correctly both with client-side routing and a non-root public URL.
       Learn how to configure a non-root public URL by running `npm run build`.
     -->
-    <title>cmd-f Portal</title>
+    <title>HackCamp Portal</title>
```
jackyzha0, line 23: nit: would export this to `utility/Constants.js@copyText`

jackyzha0 (approved):
banger work king

public/index.html
kevin-zou: Am I able to do this or would I have to use an env variable?

jackyzha0: don't think you need an env var, just add a link to the object and then reference the object here :)) just makes changing things easier for the future

kevin-zou: Might be out of my element here, but have no idea how to do this :') Would you mind providing some code snippets on how to achieve this?

jackyzha0: I think the easiest would probably to remove the definition here and use `next/head` https://nextjs.org/docs/api-reference/next/head and a Custom `App` https://nextjs.org/docs/advanced-features/custom-app

BUT this is a lot of effort for not a lot of gain so u can probs ignore

kevin-zou: OH gotcha, yeah I think I'll get this out for the time being and look into doing this for the future. Good flag though 👍

PR #301, opened 2021-11-13
Fix parser for opt-in prizing parsing

kevin-zou:
## Description
Parsing was broken for opt-in prizing, so this hopefully fixes it!


jackyzha0 (approved):
:shipit:

PR #311, opened 2021-11-18
Phoenix/part2

### LINKED ISSUE #286: Hacker App > Part 2
Based on the developer/designer selection from part 1, certain fields should be indicated as mandatory (update validation as needed):
- If developer: resume and GitHub are mandatory
- If designer: resume and portfolio are mandatory

- Q12: Update "GitHub" to "GitHub/BitBucker/GitLab Link", Update "Portfolio" to "Personal website/portfolio link", Update "LinkedIn" to "LinkedIn Link"
- Q13: Have two written questions, first one mandatory, second as a pick one of the of the following prompts
1. What should technology be used for?
2. Option 1: How would you like to challenge yourself during this hackathon? Option 2: Describe a time where you went above and beyond of your role to demonstrate leadership in a project.

phoenixthefirebird:
## Description
Updating this year's questions and required skill info

closes #286
## Other considerations
- error messages are squished between linkes without white space and seems to overflow from the parent div, I've tried some display blockl and box sizing to no avail, any help is appreciated!

### DIFF
--- a/src/components/ApplicationForm/Skills.js
+++ b/src/components/ApplicationForm/Skills.js
@@ -102,7 +102,7 @@ export default ({ errors, formInputs, onChange, role, handleResume }) => {
             <FormRow id="GitHub/BitBucket/GitLab">
               <TextInput
                 inline
-                placeholder="Optional"
+                placeholder="Input your link"
                 size="large"
                 value={formInputs.github}
                 invalid={!!errors.github}
@@ -152,7 +152,7 @@ export default ({ errors, formInputs, onChange, role, handleResume }) => {
             <FormRow id="Personal website/portfolio link">
               <TextInput
                 inline
-                placeholder="Optional"
+                placeholder="Input your link"
                 size="large"
                 value={formInputs.portfolio}
                 invalid={!!errors.portfolio}
--- a/src/components/ApplicationForm/Skills.js
+++ b/src/components/ApplicationForm/Skills.js
@@ -178,12 +178,12 @@ export default ({ errors, formInputs, onChange, role, handleResume }) => {
         <StyledTextArea
           maxLength="650"
           width="100%"
-          value={formInputs.longAnswers}
-          invalid={!!errors.longAnswers}
-          errorMsg={errors.longAnswers}
+          value={formInputs.longAnswers1}
+          invalid={!!errors.longAnswers1}
+          errorMsg={errors.longAnswers1}
           onChange={val =>
             onChange({
-              longAnswers: val,
+              longAnswers1: val,
             })
           }
         />
@@ -201,12 +201,12 @@ export default ({ errors, formInputs, onChange, role, handleResume }) => {
         <StyledTextArea
           maxLength="650"
           width="100%"
-          value={formInputs.longAnswers}
-          invalid={!!errors.longAnswers}
-          errorMsg={errors.longAnswers}
+          value={formInputs.longAnswers2}
+          invalid={!!errors.longAnswers2}
+          errorMsg={errors.longAnswers2}
           onChange={val =>
             onChange({
-              longAnswers: val,
+              longAnswers2: val,
             })
           }
         />
--- a/src/components/ApplicationForm/Skills.js
+++ b/src/components/ApplicationForm/Skills.js
@@ -170,16 +170,33 @@ export default ({ errors, formInputs, onChange, role, handleResume }) => {
 
       <FormSpacing>
         <QuestionHeading>question 13</QuestionHeading>
-        <SubHeading color="primary">
-          Answer one of the two questions:
+        <SubHeading color="primary">Two written questions:</SubHeading>
+        <SubHeading size="1.25em">
+          1. What should technology be used for?
+          <Required />
+        </SubHeading>
+        <StyledTextArea
+          maxLength="650"
+          width="100%"
+          value={formInputs.longAnswers}
+          invalid={!!errors.longAnswers}
+          errorMsg={errors.longAnswers}
+          onChange={val =>
+            onChange({
+              longAnswers: val,
+            })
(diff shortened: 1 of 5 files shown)


src/components/ApplicationForm/Skills.js
```diff
@@ -83,37 +83,71 @@ export default ({ errors, formInputs, onChange, role, handleResume }) => {
...
-                errorMsg={errors.portfolio}
-                onChange={e =>
-                  onChange({
-                    portfolio: e.target.value,
-                  })
-                }
-              />
-            </FormRow>
+            <>
+              <FormRow id="Personal website/portfolio link">
```
kevin-zou, line 87: For `id` values can we keep it to alphanumeric characters and use hyphens or underscores to break words? Don't think we're supposed to have whitespace in `id`

Docs: https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/id

src/components/ApplicationForm/Skills.js
```diff
@@ -83,37 +83,71 @@ export default ({ errors, formInputs, onChange, role, handleResume }) => {
...
+                  invalid={!!errors.portfolio}
+                  errorMsg={errors.portfolio}
+                  onChange={e =>
+                    onChange({
+                      portfolio: e.target.value,
+                    })
+                  }
+                />
+              </FormRow>
+              <FormRow id="GitHub/BitBucket/GitLab">
```
kevin-zou, line 102: Don't think we can use `/` in `id` either

src/components/ApplicationForm/Skills.js
```diff
@@ -83,37 +83,71 @@ export default ({ errors, formInputs, onChange, role, handleResume }) => {
...
-                errorMsg={errors.github}
-                onChange={e =>
-                  onChange({
-                    github: e.target.value,
-                  })
-                }
-              />
-            </FormRow>
+            <>
+              <FormRow id="GitHub/BitBucket/GitLab">
```
kevin-zou, line 120: Same as above

src/utility/Validation.js
```diff
@@ -71,6 +72,13 @@ const validateTrueFunction = thing => {
   }
 }
 
+const mandatoryURLFunction = thing => {
+  return {
+    error: thing ? true : !validateURL(thing),
```
kevin-zou, line 77: Sorry having trouble wrapping my head around line 77 - so `thing` here is the URL, and if it exists at all it will return saying that an error exists?

kevin-zou (commented):
Left some feedback! Great first pass though

panjenny0 (commented):
Agree with @kevin-zou's comments!

Regarding your issue:

> error messages are squished between linkes without white space and seems to overflow from the parent div, I've tried some display blockl and box sizing to no avail, any help is appreciated!

You could fix this in the `TextInput` component itself. A quick way is to perhaps reduce the `margin` on `TextInputContainer` and set `margin-bottom` to < 1em. This might change how this component is rendered on other pages tho so it's ok if we just leave it.

Also your changes in part[0-2] exposed an existing issue with how the application object is fetched so I opened a PR #313 to handle this - just a FYI for others who are testing your change

src/components/ApplicationForm/Skills.js
phoenixthefirebird: This id is also use to display the field value tho, I was actually considering renaming the property from id to something else like "fieldName". In particular the id field wasn't used in css styling

src/components/ApplicationForm/Skills.js
phoenixthefirebird: see above

src/components/ApplicationForm/Skills.js
phoenixthefirebird: ^

src/utility/Validation.js
phoenixthefirebird: Thank you for catching that! I've definitely swapped the two terms in the ternary statement, it should validate the URL when there is a thing and set error to true when the input is empty (will add empty string comparison as well)

phoenixthefirebird:
> Agree with @kevin-zou's comments!
> 
> Regarding your issue:
> 
> > error messages are squished between linkes without white space and seems to overflow from the parent div, I've tried some display blockl and box sizing to no avail, any help is appreciated!
> 
> You could fix this in the `TextInput` component itself. A quick way is to perhaps reduce the `margin` on `TextInputContainer` and set `margin-bottom` to < 1em. This might change how this component is rendered on other pages tho so it's ok if we just leave it.
> 
> Also your changes in part[0-2] exposed an existing issue with how the application object is fetched so I opened a PR #313 to handle this - just a FYI for others who are testing your change

Yes synching with the template would make the code so much cleaner!

panjenny0 (dismissed):
🥇

kevin-zou (dismissed):
Thanks for incorporating the feedback so quickly! Lgtm 🚀

src/utility/Validation.js
```diff
@@ -172,6 +170,7 @@ const validators = {
     resume: noInvalidResumeFunction,
     portfolio: isDesigner ? mandatoryURLFunction : optionalURLFunction,
     linkedin: optionalURLFunction,
+    hackathonsAttended: noEmptyFunction,
```
kevin-zou, line 173: Why does moving this under the `skills` field fix this? Just want to check because the current template still has `hackathonsAttended` under `basicInfo` in `Constants.js`

kevin-zou (approved):
Lgtm :)) thanks for digging into the error

src/utility/Validation.js
kevin-zou: Resolved through Slack, `hackathonsAttended` was moved to Skills section

PR #313, opened 2021-11-19
HackerApplicationContext: Sync user object with template upon fetch

### LINKED ISSUE #312: Hacker App > Sync existing user object with template
Adding new fields into the `HACKER_APPLICATION_TEMPLATE` (i.e. new `formInput.<field>`) when a user's document already exists in the collection will make the page trying to render the new field crash.

This is because the user's application object does not have the new field and the fields are not created when we fetch the user's object.

We should handle this in the HackerApplicationContext instead of individual form containers.

Screen Shot 2021-11-18 at 11 27 08 AM

panjenny0:
## Description

See description of bug in #312 

Add function to recursively construct missing fields in hacker object synced with the `HACKER_APPLICATION_TEMPLATE`
- Also works for new field within nested obj

## Testing instructions
1. go to non-prod db
2. remove some fields from your own application object
3. reload page
4. page should not crash with error similar to #312 

Closes #312

### DIFF
--- a/src/utility/HackerApplicationContext.js
+++ b/src/utility/HackerApplicationContext.js
@@ -9,8 +9,10 @@ import {
 } from './firebase'
 import firebase from 'firebase/app'
 import Spinner from '../components/Loading'
-import { ANALYTICS_EVENTS } from './Constants'
+import { ANALYTICS_EVENTS, HACKER_APPLICATION_TEMPLATE } from './Constants'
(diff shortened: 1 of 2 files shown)


phoenixthefirebird (approved):
wow I didn't know we already the helper function needed to synch this up, lgtm!

kevin-zou (approved):
lgtm! the utility function is super cool 🚀

PR #318, opened 2021-11-26
small fixes

naijwu:
## Description

**All changes**
- `src\containers\Application\Part0.js` Replaced 'Save & Logout' to back button 
- `src\containers\Application\Part0.js` Added back button
- `src\components\ApplicationForm\Skills.js` Changed formatting of Q14:
choose2

- `src\App.js` Added ternary in PageRoute component: if portal is in application state, redirect to /application rather than display page
- `public\index.html` Changed title to "nwHacks Portal"
- `public\favicon.ico` Changed favicon to nwHacks 2022
- `src\utility\Constants.js` changed hackathonName to 'nwHacks 2022' **_!!!_** still need to change other constants

## Other considerations

- Add flag for when portal goes from application -> day-of portal?


src/App.js
```diff
@@ -38,9 +38,11 @@ const notifyUser = announcement => {
 }
 
 const PageRoute = ({ path, children }) => {
+  const ONLY_APPLICATION = true // CHANGE: when not application
```
panjenny0, line 41: Can we actually move this to `Constants.js`? Ideally we would have these configs in one place.

panjenny0 (commented):
This is great! Thank you for making these changes.
While you're at it, can you also change the `DB_HACKATHON` collection to `nwHacks2022`? 
https://github.com/nwplus/portal/blob/eff2864d93f34b0012863008f00f5b29cdbfdd89/src/utility/Constants.js#L4

Everything should still work after switching to the new collection but let me know if any unexpected issues come up!

src/App.js
naijwu: OKay okay of course sounds good! thank you!

panjenny0 (approved):
Lgtm!

PR #324, opened 2021-12-06
debounce, renaming, other final fixes

naijwu:
## Description

- Changed "North American Indigenous" to Indigenous (First Nations, Métis, Inuk/Inuit)
- Got rid of “Please include this field” error message when we go over the character limit, character limit continues to show and error message is replaced by "Sorry! It looks like your answer is x character(s) over the limit."
- Changed phone number verification regex match so it's exactly the "+X XXX XXX XXXX" Format
- Other text field is shown when selecting "Other" in question 8; stored in Firestore as `otherEducationLevel`
- Didn't move question 12 to Part 1: kept it in Part 2 but if hackathon years of experience = 0, made would-be mandatory fields optional
- Enabled debounce (w/ 1 second throttle) for schools dropdown. Much less lag :D

## Other considerations

- For phone number, what about people from different area codes?

### DIFF
--- a/src/components/ApplicationForm/BasicInfo.js
+++ b/src/components/ApplicationForm/BasicInfo.js
@@ -36,7 +36,6 @@ const educationOptions = [
 ]
 
 const graduationOptions = [
-  { value: 2019, label: '2019' },
   { value: 2020, label: '2020' },
   { value: 2021, label: '2021' },
   { value: 2022, label: '2022' },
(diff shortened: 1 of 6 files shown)


anneguo3 (changes requested):
Looks amazing 🎉 ❤️  just a couple last small things 
- can we default enable debounce on the generic input field? - or at least add them for the long answer questions? 
- remove "2019" as an option from the "Grad Year" field (question 9)
- for phone number, I believe we need to handle for when the international prefix can be more than one digit, I'm totally okay if we go for a more relaxed rule with something like [this](https://stackoverflow.com/a/33561517)

kevin-zou (approved):
holy, great work ❤️ everything looks good to me, thanks so much for being on point with this!

anneguo3 (approved):

PR #326, opened 2021-12-09
Debounce updating application

kevin-zou:
## Description
Follow up to #324. Text input fields should no longer be laggy. Thanks to the excellent code from @jackyzha0 we have a debounce utility now!

Previously we had a `setTimeout` that would send updates to Firebase every 30s if the application had changed since. Now what happens is whenever the local application object gets updated, a debounced update call is made. If no other changes are made to the application object in the next 30s, it will send an update to Firebase.

This prevents us from checking code every 30s and instead triggers when people are actively working on their application.

## Other considerations
Noticed that there was some really weird styling for specific viewports and states (screenshot below).
image

After:
image


naijwu (approved):
Looks so beautiful kevin and jacky legends

PR #335, opened 2021-12-30
Portal flags

anson1014:
## Description

- removed hardcoded use of `ONLY_APPLICATION` from `/constants.js` and instead sets the value depending on what `getLivesiteDoc` returns for the `applicationsOpen` value.
- `applicationsOpen == true` -> redirect to application page
- `applicationsOpen == false` -> redirects to day-of pages

## Other considerations

Honestly not sure if I'm doing this right, or if there are other instances where the `ONLY_APPLICATION` constant is used. But `App.js` was all that I could find.

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -45,8 +45,6 @@ const PageRoute = ({ path, children }) => {
     return unsubscribe
   }, [setLivesiteDoc])
 
-  console.log(livesiteDoc.applicationsOpen)
-
   return (
     <Route path={path}>
       {livesiteDoc.applicationsOpen ? <Redirect to="/application" /> : <Page>{children}</Page>}
--- a/src/App.js
+++ b/src/App.js
@@ -38,16 +38,16 @@ const notifyUser = announcement => {
 }
 
 const PageRoute = ({ path, children }) => {
-  const [livesiteDoc, setLivesiteDoc] = useState(false)
+  const [livesiteDoc, setLivesiteDoc] = useState(null)
 
   useEffect(() => {
     const unsubscribe = getLivesiteDoc(setLivesiteDoc)
     return unsubscribe
-  }, [setLivesiteDoc])
+  }, [])
 
   return (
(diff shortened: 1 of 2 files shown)


src/App.js
```diff
@@ -44,9 +38,16 @@ const notifyUser = announcement => {
 }
 
 const PageRoute = ({ path, children }) => {
+  const [livesiteDoc, setLivesiteDoc] = useState(false)
+
+  useEffect(() => {
+    const unsubscribe = getLivesiteDoc(setLivesiteDoc)
+    return unsubscribe
+  }, [setLivesiteDoc])
```
kevin-zou, line 46: I'm not too familiar with the `useEffect` hook, but does this change on `setLivesiteDoc` or should it change on `livesiteDoc` changing?

src/App.js
```diff
@@ -44,9 +38,16 @@ const notifyUser = announcement => {
 }
 
 const PageRoute = ({ path, children }) => {
+  const [livesiteDoc, setLivesiteDoc] = useState(false)
```
kevin-zou, line 41: Nit: I would set this to `null`, then use an optional chaining operator on line 50 to prevent "cannot read property of null" errors - lmk though! Something about having the default state for an object be `false` seems off to me, but it doesn't seem to have any functional impact either way

kevin-zou (approved):
AYYY let's get it, thanks for the quick turnaround. Tested with dev Admin and it works perfectly 👌

src/App.js
anson1014: good catch :)

src/App.js
anson1014: you're right, it's not supposed to be there. It's only supposed to fetch the `LivesiteDoc` on the initial render.

kevin-zou (approved):
thanks for the quick changes! lgtm :)

naijwu (approved):

PR #336, opened 2022-01-04
Project submission form

### LINKED ISSUE #333: Project submission form
Update the current submission page to be a fully fledged form that has the following fields:
- Project name (char limited)
- Team members (name + email)
- Short description (char limited)
- Description
- Links*

*There can be a lot of different links (eg. YouTube demo link, deployed link, code, design docs), so could we either:
1) Have a plus icon that makes a new field, and a delete icon that removes a field, or
2) Just have 5 fields from the start (all of them which are not required)

The form should be fully validated as well, using functions from `Validation.js` or making new ones there.

kevin-zou:
## Description
Closes #333. This PR introduces the frontend for a project submission system. It replaces the content on the Devpost linking page with this form.

### Default state
image
image

### Error states
image
image

### DIFF
--- a/src/components/Input/TextInput.js
+++ b/src/components/Input/TextInput.js
@@ -43,6 +43,11 @@ const TextInputBox = styled.input.attrs({
     `}
 `
 
+// max-width: fit-content to make the error message only span the text input width
+const StyledErrorMessage = styled(ErrorMessage)`
+  max-width: fit-content;
+`
+
 export const TextInput = ({
   className,
   invalid,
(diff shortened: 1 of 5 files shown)


ianmah (approved):

anneguo3:
Can we have prizes as a multi-select like how we do select for ethnicities in application? The spacing is just a lil off

kevin-zou:
> Also was there a reason behind having a max 240 characters for the description?

Hmmm nope, I think I saw it somewhere before but besides that it's pretty arbitrary. We should change this though I think, especially considering that this is supposed to replace Devpost descriptions. Wasn't sure how to "limit" it so that people can't post huge things (or do we not mind?)

kevin-zou:
@anneguo3 how do these look?

image
image

Issue #330, opened 2021-12-29
Add categories to sidebar


kevin-zou:
The sidebar on Portal is being reworked to have categories for the links. The categories and their links (in order) are listed below. An arrow `->` means the page exists and is being renamed.

**General**
Quicklinks -> Getting Started
Dashboard -> Home
Schedule
Sponsors

**Tools**
Project Gallery (new)
Submission -> Project Submission
Judging -> Peer Judging
Judging Admin (organizer only)

**Information**
Info Package (new)
Judging (new)
FAQ

The headers should be in a font that is smaller and lighter in the color than the links. If you need design guidance feel free to ask for it in the slack channel!

naijwu:
Is the Judging link under the Information category new? Or would it be the rename of "Judging Admin"?

kevin-zou:
Hmm good question, I think it is but @anneguo3 do you know?

kevin-zou:
Just asked Anne, I'll update the description to match!

PR #338, opened 2022-01-06
Sidebar Categories

naijwu:
## Description

Added categories to sidebar based on #330 and applied styling based on [this figma](https://www.figma.com/file/bSh3usa74gfv1VvHH7Z4bf/nwPlus-Live-Site-Design-Direction?node-id=1351%3A218)

**not production ready; look at considerations below (Broken links)**

sidebar

#### Broken links:
- Tools > Project Gallery
- Information > Info Package
- Information > Judging

#### Styling considerations:
- Hard-coded some colours _for now_ (instead of editing `ThemeProvider`) to not break styling on other elements

### DIFF
--- a/src/components/Sidebar.js
+++ b/src/components/Sidebar.js
@@ -18,7 +18,6 @@ const SidebarContainer = styled.div`
   transition: opacity 1s ease-out;
   z-index: 1;
   background: #051439;
-  text-transform: uppercase;
   ${p => p.theme.mediaQueries.mobile} {
     ${p => (p.showMobileSidebar ? 'visibility: visible' : 'visibility: hidden; display: none')};
   }
@@ -64,6 +63,7 @@ const ItemsContainer = styled.div`
 `
 
 const StyledA = styled(A)`
+  text-transform: uppercase;
   display: block;
   font-weight: bold;
   padding: 1em 50px;
@@ -71,10 +71,12 @@ const StyledA = styled(A)`
   color: ${p => (p.theme.name !== 'cmdf' && p.selected ? '#051439' : p.theme.colors.highlight)};
   ${p => p.selected && `background: #FFB72C;`}
   &:hover {
+    color: #ffffff;
     background: ${p => p.theme.colors.secondaryBackgroundTransparent};
     border-bottom: none;
   }
   &:focus {
+    color: #ffffff;
     background: ${p => p.theme.colors.secondaryBackgroundTransparent};
     border-bottom: none;
   }
@@ -83,7 +85,7 @@ const StyledA = styled(A)`
 const LiveDot = styled.span`
   height: 10px;
   width: 10px;
-  background-color: ${p => p.theme.colors.background};
+  background: ${p => p.theme.colors.background};
   border-radius: 50%;
   margin: 0 7px 0 4px;
   display: inline-block;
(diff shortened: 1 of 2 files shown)


src/components/Sidebar.js
```diff
@@ -130,12 +142,30 @@ export default ({
...
+  const linksDict = ['General', 'Tools', 'Information', '']
   const links = [
-    { location: '/', text: 'DASHBOARD' },
-    { location: '/schedule', text: 'SCHEDULE' },
-    { location: '/quicklinks', text: 'QUICKLINKS' },
-    { location: '/faq', text: 'FAQ' },
-    { location: '/sponsors', text: 'SPONSORS' },
+    // General
+    [
+      { location: '/quicklinks', text: 'Getting Started' },
```
kevin-zou, line 149: What do you think of making `links` a dictionary where the keys are the categories (eg. "General", "Tools")? The values in this case would just be the array of link objects as you have them.

Imo would make the code below cleaner (replaces `links[1].push(...)` with `links.tools.push(...)` which is a _bit_ more descriptive, but this is more of a suggestion than anything!

src/components/Sidebar.js
```diff
@@ -149,27 +179,51 @@ export default ({
...
+              </Link>
+            )
+          })}
+        </>
+      )
+    )
+
+    setRenderLinks(renderArray)
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [links])
```
kevin-zou, line 225: Previously toggling a flag in Admin would update the sidebar without having to refresh. Wondering:
1. Is this behaviour still the same? and if so,
2. Is the reason the code here was taken out from the JSX because we're changing the data structure of the links?

src/components/Sidebar.js
```diff
@@ -116,10 +120,18 @@ const StatusText = styled.div`
 
 const SponsorLogo = styled.img`
   display: block;
-  margin: 1em 0 0 60px;
+  margin: 1em 0 0 50px;
```
kevin-zou, line 123: Random request, but since you're already working in sidebar would you mind fixing the margin on the "Logout" button as well? It seems like it's not aligned

src/components/Sidebar.js
```diff
@@ -14,10 +14,10 @@ import { getSponsors } from '../utility/firebase'
 const SidebarContainer = styled.div`
   min-width: 275px;
   min-height: 100%;
-  border-right: 1px solid ${p => p.theme.colors.border};
+  // border-right: 1px solid ${p => p.theme.colors.border};
   transition: opacity 1s ease-out;
   z-index: 1;
-  background: ${p => p.theme.colors.background};
```
kevin-zou, line 20: Can you keep this line but comment it out so we know what it was like before? If we could do this for the other replaced CSS values that would be fantastic

src/theme/ThemeProvider.js
```diff
@@ -44,19 +44,19 @@ const nwTheme = {
...
   colors: {
-    background: '#2C2543',
-    card: '#2C2543', // todo: update to proper colour
+    background: 'linear-gradient(to bottom, #52999F, #051439)',
+    card: '#2C2543', // TODO: update to proper colour
     border: 'rgba(255, 255, 255, 0.3)',
-    secondaryBackgroundTransparent: '#1D1B24bb',
-    secondaryBackground: '#1D1B24',
+    secondaryBackgroundTransparent: '#FFB72Cbb',
+    secondaryBackground: '#1D1B24', // before: #1D1B24
```
kevin-zou, line 51: Nit: this color didn't change

kevin-zou (commented):
Great first pass! Left some comments, lmk what you think/if you have any questions!

src/components/Sidebar.js
naijwu: I used a design pattern I was familiar w/ for rendering array data BUT how it was before is much better so I've since changed the code. By extension, previous behavior should stay the same

src/components/Sidebar.js
naijwu: 100%. I was boosted when I committed this code

kevin-zou: no you were not king 😤

kevin-zou (approved):
LGTM! thanks for taking care of the theme changes as well!

Issue #345, opened 2022-01-08
Schedule improvements


kevin-zou:
### Problems
- currently schedule will break its layout if there are more than two events happening at the same time
- as the hackathon progresses, it's kind of annoying to scroll further down (schedule always begins at start date + time)

### Developing with schedule
The easiest way to get dummy events to show up is through locally running Admin and:
- going to Livesite settings and setting "nwHacks2022 Start Time" and "nwHacks2022 End Time" to whatever dates you want
- go to the dashboard for nwHacks2022 and add three events happening at the same time (on the dates you set previously)

### Requirements
- the schedule page should be responsive no matter how many events there are
  - my idea for this is to add a horizontal overflow scroll to the container, but if you have another idea let me know!
- schedule scrolls to current date + time (this is more a QOL thing, so if you feel like you're getting nowhere it's fine to leave out)

PR #347, opened 2022-01-09
scrollable schedule implementation

dchen150:
## Description

#345 
image

## Other considerations

NOTE: I think ideally the time stamps stay while the rest of the schedule is scrollable - I tried implementing this but it was an absolute CSS hell. Run into problems with the dashed lines and the columns not showing up.

### DIFF
--- a/src/components/Schedule/Schedule.js
+++ b/src/components/Schedule/Schedule.js
@@ -119,26 +119,3 @@ export default ({ events, hackathonStart, hackathonEnd }) => {
     </OverflowContainer>
   )
 }
-
-{
-  /* <OverflowContainer header="Day-Of-Events Schedule">
-  <TagLegend />
-
-  <ScheduleFlexContainer>
-    <TimelineColumn
-      hackathonStart={hackathonStart}
-      duration={durationOfHackathon}
-      numCols={schedule.length}
-    />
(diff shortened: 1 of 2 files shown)


src/components/Schedule/Schedule.js
```diff
@@ -6,9 +6,35 @@ import { TimelineColumn } from './Timeline'
 import { TagLegend } from './Tag'
 import Event from './Event'
 
+const ScrollableContainer = styled.div`
+  overflow-x: scroll;
+  overflow-y: hidden;
+  transform: rotateX(180deg);
```
kevin-zou, line 12: Nit: correct me if I'm wrong, but the rotation on line 12 is so that the scrollbar shows up on the top of the container and not the bottom? If this is the case maybe we can leave a quick comment on this above the start of this component

kevin-zou (approved):
lgtm! when I was playing around with it the scrollbar seems to show up even if there is enough room. not a big deal but if we can find a solution for this that would be fantastic
image

src/components/Schedule/Schedule.js
dchen150: yep that's right - i wasn't able to find a way to make the scroll bar on top other than doing the rotation. i'll leave a comment :)

PR #358, opened 2022-01-11
Hacker only auth

kevin-zou:
## Description
With us bringing project submissions to Portal, it makes sense to allow only accepted hackers to the submission page. I've also set up this sort of authentication for the peer judging page. Not sure if there's more work to be done here, but it seems pretty solid so far.

Right now behaviour is if you are anything but accepted, on login it will bring you to the application review page where it tells you your application status. If you navigate to the home page and try visiting these auth'd pages you will get redirected immediately to the index page.

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -67,7 +67,7 @@ const AuthPageRoute = ({ path, children }) => {
   }
   return (
     <Route path={path}>
-      {user.status === APPLICATION_STATUS.accepted ? <Page>{children}</Page> : <Redirect to="/" />}
+      {user?.status === APPLICATION_STATUS.accepted ? <Page>{children}</Page> : <Redirect to="/" />}
     </Route>
   )
 }
--- a/src/App.js
+++ b/src/App.js
@@ -55,9 +55,21 @@ const PageRoute = ({ path, children }) => {
   )
 }
 
+// Authenticate for only applicants that have been accepted
 const AuthPageRoute = ({ path, children }) => {
-  const { isAuthed } = useAuth()
-  return <Route path={path}>{isAuthed ? <Page>{children}</Page> : <Redirect to="/login" />}</Route>
+  const { isAuthed, user } = useAuth()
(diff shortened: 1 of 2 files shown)


src/App.js
```diff
@@ -55,9 +55,21 @@ const PageRoute = ({ path, children }) => {
   )
 }
 
+// Authenticate for only applicants that have been accepted
 const AuthPageRoute = ({ path, children }) => {
-  const { isAuthed } = useAuth()
-  return <Route path={path}>{isAuthed ? <Page>{children}</Page> : <Redirect to="/login" />}</Route>
+  const { isAuthed, user } = useAuth()
+  if (!isAuthed) {
```
jackyzha0, line 61: this is probably a redundant check but should we include `!(isAuthed || user)` so L70 doesn't ever become undefined?

src/utility/Auth.js
```diff
@@ -44,7 +44,7 @@ export function AuthProvider({ children }) {
     analytics.logEvent(ANALYTICS_EVENTS.Logout, { userId: user.uid })
     await firebase.auth().signOut()
     setUser(null)
-    setLocation('/login')
+    setLocation('/')
```
jackyzha0, line 47: why this change?

kevin-zou: Thought it was a better experience than the current one. On logout you're redirected to `/login`, but it's a "dead end" page since you can't go back in the browser and there's no button for home.

src/App.js
kevin-zou: Would changing line 70 to be `user?.status` work? (If I'm understanding your comment correctly) I feel a bit weird about including a check on `isAuthed` because if the user isn't auth'd then we'd just send them to `/`, which isn't very helpful.

jackyzha0: ah yes that would work too!

jackyzha0 (approved):
lgtm king

PR #363, opened 2022-01-13
Clear missing submittedProject

ianmah:


kevin-zou (approved):
Maybe we should leave a comment about why we do this, but otherwise lgtm

PR #366, opened 2022-01-13
autofill and validate self submission

anson1014:
## Description

- autofills the currently auth'd user when there is no pre-existing project submission
- validates that the currently auth'd user must be in their own submission

### DIFF
--- a/src/containers/SubmissionLink.js
+++ b/src/containers/SubmissionLink.js
@@ -28,6 +28,7 @@ const tempProject = {
 export default ({ user, refreshCallback }) => {
   const [project, setProject] = useState(tempProject)
   const [isSubmitting, setSubmitting] = useState(false)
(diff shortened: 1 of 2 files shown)


src/containers/SubmissionLink.js
```diff
@@ -40,7 +41,17 @@ export default ({ user, refreshCallback }) => {
...
         const projectDoc = await projectsRef.doc(projectId).get()
         setProject({ ...projectDoc.data(), uid: projectDoc.id })
+      } else {
+        let autoFill = [
+          {
+            name: userData.basicInfo.firstName + ' ' + userData.basicInfo.lastName,
+            email: userData.basicInfo.email,
+          },
+        ]
+        console.log(autoFill)
```
ianmah, line 51: 🔪🔪🔪

PR #367, opened 2022-01-13
autofill and validate self submission

anson1014:
## Description

- autofills the currently auth'd user when there is no pre-existing project submission
- validates that the currently auth'd user must be in their own submission


ianmah (approved):
lgtm

PR #369, opened 2022-01-13
add validation for members

vincentchiang1:
## Description

* Check if user actually exists and is an accepted hacker
* Check if user has a project already


kevin-zou (approved):
lgtm 🚀

PR #370, opened 2022-01-13
Leave button

phoenixthefirebird:
alt text

## Description
`Leave Project` button is implemented
- remove project id from user object in db
- if user is no the only team member, remove user from the project
- if user is the only team member, delete the project 

## Other considerations
- let me know if the `window.location.reload()` is a good idea and if I should replace it with something else, also added it to when user click the `Save` button


kevin-zou (approved):
Lgtm! 🔥

PR #373, opened 2022-01-14
Submission improvements

kevin-zou:
## Description
- when saving it doesn't throw an error saying that you belong to another project
- handle if an email is not found
- only update Firebase project document when all members are valid
- support `lastEditedBy`
- added comments

image

### DIFF
--- a/src/containers/SubmissionLink.js
+++ b/src/containers/SubmissionLink.js
@@ -73,63 +73,100 @@ export default ({ user, refreshCallback }) => {
     delete projectSubmission.uid
     if (projectId) {
       try {
-        await projectsRef.doc(projectId).update(projectSubmission)
         // TODO: Determine if these emails are "new" emails
-        // TODO: Check that the person doesn't already have a project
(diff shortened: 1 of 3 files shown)


src/containers/SubmissionLink.js
```diff
@@ -73,63 +73,104 @@ export default ({ user, refreshCallback }) => {
...
               } else if (res.docs[0].data().submittedProject) {
                 setError(
                   new Error(member.email + ' is already part of a different project submission.')
                 )
-              } else
+              } else {
+                // On first valid member, create a project to be used
+                if (!project) {
+                  project = await createProject(user.email, projectSubmission)
+                }
```
kevin-zou, line 149: This is to prevent us from making a project in Firebase just to have it orphaned because the email was not valid

anson1014 (approved):
i think it LGTM!

ianmah (approved):

phoenixthefirebird (commented):
left some comments, sorry for the pending!

PR #378, opened 2022-01-14
count assigned on creation

vincentchiang1:
## Description

* add missing count assigned field when we create a project


kevin-zou (approved):
ayyyyy

ianmah (approved):

PR #386, opened 2022-01-15
Implement success toast

kevin-zou:
## Description
Rewrote the existing ErrorBanner to be Toast, and added ability to specify whether it's a success or error toast. Currently defaults to Toast so that existing usage of the component doesn't change (this has been checked)

With this done, I've now set the toast to show up on successful actions in the project submission form, and it'll show for 3 seconds before the page is reloaded automatically.

image


phoenixthefirebird (approved):
LGTM! Thank you for polishing this up it's very sleek

PR #389, opened 2022-01-15
Handle another case + copy change

kevin-zou:
## Description
Handle the case where a user tries to make a new project submission with invalid users.

Change copy for Microsoft prize.


phoenixthefirebird (approved):
lgtm 🚀

PR #394, opened 2022-01-15
ui for peer judging

phoenixthefirebird:
## Description
- make the sizing and layout of the peer judging to be the same as what's in project gallery
- break word in long description and add a scroll bar


kevin-zou (approved):
Very nice! Like the scrollbar in case the text goes over 👍

PR #400, opened 2022-01-16
filter projects using dropdown and fetch prizes from firebase

vincentchiang1:
## Description

image


kevin-zou (approved):
ayyy let's get it 👌

PR #408, opened 2022-01-18
Add mentor nominations column to CSV

kevin-zou:
## Description
As per logs request so they can manually count the number of mentor nominations. Example of exported CSV below:

image


anneguo3 (approved):
lgtm! 🚢

PR #422, opened 2022-03-02
Charity dropdown

phoenixthefirebird:
## Description
Added Dropdown for charity choice
image

## Other considerations
The sponsorprizes are currently crashing the app due to being undefined, so I added the conditional render for sponsor prizes. We should fix that


kevin-zou (changes requested):
Looks good! the sponsor bug should be fixed on `main` already, so that should resolve after you address the merge conflicts.

Can you also add `charityChoice` to the `onSubmit` call starting on line 234? Otherwise the data won't get written to the db

phoenixthefirebird:
> Looks good! the sponsor bug should be fixed on `main` already, so that should resolve after you address the merge conflicts.
> 
> Can you also add `charityChoice` to the `onSubmit` call starting on line 234? Otherwise the data won't get written to the db

my bad it's now fixed!

kevin-zou (approved):
Lgtm! Looks like there's a merge conflict with the code upstream, but once you fix that you're good to go! 🚢

PR #427, opened 2022-03-04
Update styles for card and schedule

kevin-zou:
## Description
I'd like to think my code is usually pretty good, but I am sure this time it's not. I just feel like there was a bunch of different stuff going on here and it feels messy. I hate it here.

### Removing cmd-f dependent styles
So for last year's cmd-f flavour for Portal we put a bunch of conditional styling that read something like
```
p => p.theme.name === 'cmdf' ? <cmd-f styles> : <regular styles>
```

We did it this way before because there were neat images and stuff, but this year we don't have that. I removed the instances of this that I could find because it was really messing with our "regular" cmd-f theme that we have this year.

### Updating card styles (I sure hope so)
Basically I'm putting blanket styling on how cards should look in Portal. I thought it was strange that cards depended on `secondaryBackground`. Thus I've introduced `card` and `cardText`. All Cards will now have a bg of `card` and their text will be color `cardText`.

image

### Schedule
We're going with a bit of a different schedule design this year. The problem I was running into was that the styles for the schedule page was being overwritten by the styles from `CardLike` in `Common.js`. My solution is to copy `CardLike` into `Schedule.js` (removing the styles that I don't want).

I left a TODO so that we can come back to this PR and see that I was very dumb and we should go back to how things were before. If anybody has a better solution here PLEASE let me know, but I feel like I'm in too deep.

image


src/components/Common.js
```diff
@@ -1,66 +1,38 @@
...
     padding: 1em;
     margin: 0.75em 0;
   }
+
+  & > h1,
+  h2,
+  h3,
+  p {
+    color: ${p => p.theme.colors.cardText};
+  }
```
kevin-zou, line 21: This is to force all text in cards to be color `cardText`

vincentchiang1 (approved):
:)

PR #428, opened 2022-03-04
new home styles

vincentchiang1:
## Description

- new leaf background
- update the card color for the Twitch stream embed

image


kevin-zou (commented):
Can we actually change `card` to be `#2C3933`? Requested by design

kevin-zou (approved):

PR #431, opened 2022-03-05
Hardcode sidebar sponsor logos

kevin-zou:
## Description
The bg of the sidebar is dark so we want light variants of logos. The bg of the Sponsors page is light so we want dark variants of logos there. Since we don't have light/dark variants we should just hardcode this for cmd-f and call it a day :))

image


vincentchiang1 (approved):
:D

PR #433, opened 2022-03-05
cmd-f 2022 nits + schedule scroll

kevin-zou:
## Description
Added a case for devpost links in project views so that it doesn't show up as a long ugly link
image

Also finally added a scroll to current time in schedule :))) lmk if you can think of a cleaner implementation, but this is what I came up with

### DIFF
--- a/src/components/Schedule/Timeline.js
+++ b/src/components/Schedule/Timeline.js
@@ -43,6 +43,7 @@ const TimelineLabel = styled.span`
 
 const CurrentTime = ({ start, duration, numCols }) => {
   const [currentTime, setCurrentTime] = useState(Date.now())
+  const [scrolled, setScrolled] = useState(false)
   const hoursBetweenNowAndStart = (currentTime - start) / 60 / 60 / 1000
   const renderCurrentTime = 0 < hoursBetweenNowAndStart && hoursBetweenNowAndStart < duration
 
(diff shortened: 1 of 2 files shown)


anneguo3 (approved):
🚢

src/components/Schedule/Timeline.js
```diff
@@ -53,6 +54,22 @@ const CurrentTime = ({ start, duration, numCols }) => {
     }
   }, [])
 
+  // Scroll to where the current time indicator is
+  // I use a timeout because for some reason window initially thinks its height is short, so
+  // the scroll maxes out at this point. Obviously we don't want to, so a small timeout here.
+  useEffect(() => {
+    if (!scrolled && hoursBetweenNowAndStart > 1) {
+      const timeout = setTimeout(() => {
+        const scrollHeight = 72 + hoursBetweenNowAndStart * HOUR_HEIGHT
```
jackyzha0, line 63: what is the 72 here?

jackyzha0 (approved):
lgtm! small question

src/components/Schedule/Timeline.js
kevin-zou: This was the margin + padding of the page container and the schedule container. Just added this so that the current time indicator isn't right at the top of the screen but has some space!

PR #435, opened 2022-03-06
More nits

kevin-zou:
## Description
- fix some copy
- charity selection was saving but not rendering on front end
- comment out buggy filter in project gallery


anneguo3 (approved):
lgtm!!

PR #441, opened 2022-10-25
Hackcamp2022 portal

naijwu:
## Description

- Changed theme to Hackcamp
- Removed sponsors from sidebar/nav


kozr (approved):
Only looked at the preview app, but metadata (title and favicon) needs to be changed.

PR #458, opened 2022-12-09
Update review card

kozr:
## Description

- Fixed a bug for displaying if someone is a first time hacker on the review page.
- Left some comments to fix things like `formInputs.skills.hackathonsAttended` for next hackathon. Right now `formInputs.skills.hackathonsAttended` is true when someone check that they are a first time hacker, and false when they're not. This doesn't make sense at all.


DonaldKLee (approved):
lgtm!

PR #460, opened 2022-12-15
Added QR Code section

naijwu:
## Description

Conditionally displays (when logged in) a QR code with user's uid as data under Announcements card in the Home page
image


kozr:
How often is this QR code used? I think it's more intuitive to find the QR code if we put on its own page (Either from a quick access button on the navbar or some place else). Let's ask design after this PR is merged.

The code looks good though.

naijwu:
@kozr  I asked Martin about this yesterday and he said we'd be getting rid of the livestream and announcements sections-- so we'll have nothing on the home page

I guess the question now is: do we rename it to "QR Code" or keep it the home page but it only has the qr code

PR #463, opened 2023-01-01
 added rsvp and dietary restriction

naijwu:
## Description

- added rsvp
- added dietary restrictions field
- added login button on side bar


naijwu:
happy new years

PR #466, opened 2023-01-01
Added safewalk option

DonaldKLee:
Added a safewalk checkbox (first time using firebase, idk if this saves to firebase or if smt will break, plz review thoroughly 💀)


DonaldKLee:
> Added a safewalk checkbox (first time using firebase, idk if this saves to firebase or if smt will break, plz review thoroughly 💀)

Update: it saves to firebase :)

PR #473, opened 2023-01-19
change handling of zero duration events

anson1014:
image

^ 0 duration events are too wide so set them to be duration 0.5

however, as a result of this, the event scheduling placement algorithm causes overlaps like below:
image
because while the duration (how height is calculated) is getting set to 0.5, latestTime incorrectly thinks it is 30 minutes earlier than it should be. latestTime is now getting set in accordance to zero duration events

what it looks like after changes:

image


kozr (approved):
I barely understand how it works but I think it makes sense, may as well merge it in early and see how prod reacts.

PR #482, opened 2023-01-24
resume download links

naijwu:
## Description

- console logs an array of resume links of applicants who checked in

## Other considerations

only creating pr to make a production env build


naijwu:
oh I must be dumb.. preview links are still in dev environments

PR #493, opened 2023-05-31
Input validation bug fix

DonaldKLee:
## Description
Error: clicking next when q7, 13 or 15 are unfilled resulted in an error

solution:
- q7: customRef={refs['graduation']} --> customRef={refs['graduation**Ref**']}
missing "Ref" at the end of the question, so could not find where to redirect the user to when the question is missing an answer
- q13: also customRef issue, removed one of the logic thingies (see commit for more detail) causing it to not chose the correct ref
- q15: no error


meleongg (approved):
lgtm, stupid bugs :(

PR #500, opened 2023-07-26
Sketch Canvas Component

lsha0730:
## Description
image

Made this input component


src/components/ApplicationForm/Skills.js
```diff
@@ -275,6 +276,24 @@ export default ({ refs, errors, formInputs, onChange, role, handleResume }) => {
             customRef={refs['longAnswers2Ref']}
           />
         </FormGroup>
+        <FormGroup>
+          <QuestionHeading>question 21</QuestionHeading>
+          <SubHeading size="1.25em">Draw your favourite character!</SubHeading>
+          <SketchCanvas
+            width={600}
```
meleongg, line 283: I don't think this is mobile responsive right now. Looking at the npm package, it looks like react-sketch-canvas supports web & mobile

src/components/Input/SketchCanvas.js
```diff
@@ -0,0 +1,237 @@
...
+
+      <ToolBar>
+        <LeftIcons>
+          <CanvasIcon
+            src={colorpickerIcon}
+            style={{
+              marginRight: 10,
+              filter: showPicker
+                ? 'invert(31%) sepia(79%) saturate(7412%) hue-rotate(241deg) brightness(84%) contrast(127%)'
+                : 'invert(72%) sepia(0%) saturate(237%) hue-rotate(137deg) brightness(86%) contrast(87%)',
```
meleongg, line 180: Just curious how you were able to generate these filter colours for the image icons?

src/components/Input/SketchCanvas.js
```diff
@@ -0,0 +1,237 @@
...
+ * @param customRef a ref directly to the canvas component
+ * @param exportAsBase64 boolean, onChange() will return a base64 string instead of svg if true
+ */
+export default ({ width, height, invalid, errorMsg, onChange, customRef, exportAsBase64 }) => {
+  const [strokeWidth, setStrokeWidth] = useState(4)
+  const [strokeColor, setStrokeColor] = useState('black')
+  const [showPicker, setShowPicker] = useState(false)
+  const canvasRef = useRef()
+
+  const [debounceTimerId, setDebounceTimerId] = useState(null)
```
meleongg, line 108: Should we could extract the debouncer logic into its own hook to make things cleaner?

src/components/Input/SketchCanvas.js
```diff
@@ -0,0 +1,237 @@
...
+  }, [])
+
+  const handleChange = e => {
+    if (!onChange) return
+
+    if (debounceTimerId) clearTimeout(debounceTimerId)
+    const newTimerId = setTimeout(async () => {
+      if (!loadedRef.current) return
+
+      if (exportAsBase64) {
```
meleongg, line 125: Not sure I follow why we want to save as SVG for the first save and then JPEG for the following saves.

meleongg (changes requested):
AMAZING work Lincoln! SO SPEEDY! I did not know these libraries existed that could do this lol

Your overall approach LGTM, just a couple additional comments: 

- mobile responsiveness -> see ThemeProvider.js for booleans you can use within styled-components 
- I think loading from Firebase would be the next step given that you haven't touched the "value" prop passed from Skills.js -> SketchCanvas.js

src/components/Input/SketchCanvas.js
lsha0730: I used a random online codepen I found
https://codepen.io/sosuke/pen/Pjoqqp

src/components/Input/SketchCanvas.js
meleongg: see useDebounce in `utilities.js` w/i this repo

src/components/Input/SketchCanvas.js
```diff
@@ -0,0 +1,237 @@
...
+  justify-content: space-between;
+  user-select: none;
+`
+/**
+ * @param width number, width of the entire component
+ * @param height number, height of the entire component
+ * @param invalid boolean, indicates dissatisfaction with current value
+ * @param errorMsg string, indicates some error
+ * @param onChange (result: string) => void, triggers with result (SVG or Base64) on change. Has debouncing.
+ * @param customRef a ref directly to the canvas component
```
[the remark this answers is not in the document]
DonaldKLee: i believe customRef is for storing the redirect hash? like if there is a required field, and the applicant doesn't fill it out, it will scroll up to that customRef

DonaldKLee (changes requested):
Overall, great work on this feature! Amazed by how fast you completed it :)

Some notes
- Clicking on "next" after drawing and then "back" resets the drawing
   - is it possible to render a image onto their canvas? if not, a warning message might be helpful to let users know that they need to restart if they leave the page. Or you can also just display their data64 image below the canvas so applicants know that they did draw something
- Can we display their drawing on the "Review your submission" page?
- Not mobile responsive as Melvin mentioned
- Are we making the drawing required? (leaving canvas empty allows us to move on)

src/components/ApplicationForm/Skills.js
michelleykim: done (i think). I didn't touch height cuz I personally think it looks better with the same height across all devices, but lmk if you think otherwise!

src/components/Input/SketchCanvas.js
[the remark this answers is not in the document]
michelleykim: done? I believe lincoln added better documentation

src/components/Input/SketchCanvas.js
michelleykim: done (i think). lmk if you want me to change more things

src/components/Input/SketchCanvas.js
michelleykim: done

PR #508, opened 2023-10-09
Remove all vaccination info from (nwHacks) application

sidyakinian:
## Description

This PR removes all references to vaccination from portal application.

## Other considerations

As a result of this change, new applicants won't have the `vaccineInfo` field in Firebase – just something to take into account with querying, future work, etc.

### DIFF
--- a/src/containers/Application/Part0.js
+++ b/src/containers/Application/Part0.js
@@ -1,15 +1,13 @@
-import React, { useState, useEffect } from 'react'
+import React, { useState } from 'react'
 import { useLocation } from 'wouter'
 import HackathonInfo from '../../components/ApplicationForm/HackathonInfo'
 import NavigationButtons from '../../components/NavigationButtons'
 import VerticalProgressBar from '../../components/VerticalProgressBar'
 import { useHackerApplication } from '../../utility/HackerApplicationContext'
-import { checkForError, validateFormSection } from '../../utility/Validation'
 
 export default () => {
-  const { application, updateApplication, forceSave } = useHackerApplication()
+  const { application, forceSave } = useHackerApplication()
   const [, setLocation] = useLocation()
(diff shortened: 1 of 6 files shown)


src/containers/Application/Part0.js
```diff
@@ -1,6 +1,6 @@
 import React, { useState, useEffect } from 'react'
 import { useLocation } from 'wouter'
-import VaccineInfo from '../../components/ApplicationForm/VaccineInfo'
+import HackathonInfo from '../../components/ApplicationForm/HackathonInfo'
```
sidyakinian, line 3: Since there is nothing about vaccination on the first page of the application anymore

src/utility/HackerApplicationContext.js
```diff
@@ -122,10 +121,6 @@ export function HackerApplicationProvider({ children }) {
   }) => {
     const mergedApp = {
       ...application,
-      vaccineInfo: {
-        ...application.vaccineInfo,
-        ...vaccineInfo,
-      },
```
sidyakinian, line 128: If an application was started with the old version of portal and finished on the new version, it won't have `vaccineInfo` field when submitted

meleongg: In this case, I think the vaccineInfo field will remain unchanged - correct me if I'm wrong though

meleongg (approved):
Went through the testing procedures highlighted [here](https://www.notion.so/nwplus/Portal-6397f4c511d04823bb45f14b3e7ae9fb?pvs=4#87b4badb803c492dbbb390eb171f13bb) with my personal email and didn't run into any issues 

image

src/utility/HackerApplicationContext.js
sidyakinian: Oh yes you're right. Thanks for the correction

PR #514, opened 2023-11-04
add charity options for HC2023

alvinkam33:
couple of notes:
- very basic implementation, used a required dropdown (not sure if we wanted this to be required field?) it saves in the database as the field 'charityChoice', lmk if there's any other action items
- charities hard coded, not sure if there was a way in admin portal to set up charities in the DB. also made the value/label the same
- if we want to add a way for people to view a description of each of the charities (e.g. a tooltip), i could do that

### DIFF
--- a/src/components/Judging/SubmissionForm.js
+++ b/src/components/Judging/SubmissionForm.js
@@ -342,14 +342,14 @@ export default ({
         <div>
           <Label color="white">Charity Choice</Label>
           <Required />
+        </div>
+        <div>
           <P>
             Every project submitted at HackCamp 2023, regardless of completion, will be eligible for
             a $10 donation to the charity of your choice from a curated list by the HackCamp team!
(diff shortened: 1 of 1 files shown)


meleongg:
Thank you for the quick PR!

- Keeping it as required sgtm
- hardcoding charities should be okay since I don't think HackCamp will be actively changing these options
- tooltip feature could be nice -> maybe you can ping HackCamp about this on slack (https://nwplus2023-2024.slack.com/archives/C056C3YN5MW/p1698391129570019)

How did you test these changes?

alvinkam33:
yeah i just made a personal acc, applied and changed my status to acceptedAndAttending and check-in to true to access the project submission form, lmk if there are any concerns with that

meleongg (approved):
Notion link works perfectly, everything lgtm

PR #526, opened 2023-11-17
Add Youtube URL to Project Submission

meleongg:
## Description

- uncommented Youtube code
- QA testing should work but pls double check 
image
image


samanthatuhseng (approved):
looks good. 
checked that it will only accept valid youtube link. Just gotta make sure hackers youtube link actually leads to their video.
Checked that it shows up properly in peer judging, judging panel in admin view, and in project view when feedback is out. 
Looks good.

PR #528, opened 2023-11-17
added date to schedule

DonaldKLee:


meleongg (approved):
lgtm but def a band-aid fix we'll need to address for nwHacks

PR #531, opened 2023-11-22
hide troll/ heart wrenching reviews

kozr:
quick hack to hide troll reviews, might hide true positives but at least it won't hurt. Better way is to not draw them from the database in the first place.

I'll revert this once hackcamp portal is over cause it's ugly

### DIFF
--- a/src/components/Judging/Submission.js
+++ b/src/components/Judging/Submission.js
@@ -1,4 +1,4 @@
-import React from 'react'
+import React, { useEffect, useState } from 'react'
 import styled from 'styled-components'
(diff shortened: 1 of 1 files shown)


meleongg (approved):
lgtm for HackCamp, definitely not a long-term fix

I'm wondering if the residual calculation code [here](https://github.com/nwplus/portal/blob/fix-dev-cicd/src/containers/JudgingPanel.js#L66) does something similar? It looks like it discards the top 2 outliers

kozr:
ok yeah sick, I think that piece of code does it well for the admin side. Not sure if it gets rid of it from the hacker side though.

PR #535, opened 2023-11-29
Convert portal back to nwHacks 2024

meleongg:
## Description

- reskinned portal back to nwHacks
- nwHacks apps are open again
- note that you'll need to set your firebase Applicant status back to "applied" or "inProgress" and set both the nested fields to false
- also includes changes to nwHacks question ordering as mentioned [here](https://www.notion.so/nwplus/nwHacks-2024-Hacker-Application-4ff6a736debb44158f15bbca26541f4e)


DonaldKLee (approved):
lgtm!

Nitpicks:
- after submitting your application, the complete registration button is still visible
image


- there is suppose to be a coloured dot near "live", but the dot is the same colour as the background
image

- are we still including the draw a picture question that Lincoln made?
- colours aren't accurate on the hacker side (schedule page, faq, etc), I assume there'll be a new reskin in Dec and that this one is only for application?

meleongg:
- the Complete Your Registration button is greyed out so clickers can't click on it
- fixed live dot
- I don't think it's fully finished LOL

PR #545, opened 2023-12-07
Pull schedule dates from Firebase

sidyakinian:
## Description

Pulls schedule dates from Firebase instead of having them hardcoded.

## Other considerations

Changed to short date format so it's a bit easier to see the day number (see screenshot), if you think a different format is better lmk

### DIFF
--- a/src/containers/Schedule.js
+++ b/src/containers/Schedule.js
@@ -11,10 +11,10 @@ export default () => {
 
   useEffect(() => {
     const unsubscribe = livesiteDocRef.onSnapshot(doc => {
-      const d = doc.data()
-      if (d) {
-        setStart(new Date(d.hackathonStart))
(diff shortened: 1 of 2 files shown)


src/containers/Schedule.js
```diff
@@ -11,10 +11,10 @@ export default () => {
...
   useEffect(() => {
     const unsubscribe = livesiteDocRef.onSnapshot(doc => {
-      const d = doc.data()
-      if (d) {
-        setStart(new Date(d.hackathonStart))
-        setEnd(new Date(d.hackathonEnd))
+      const data = doc.data()
+      if (data) {
+        setStart(new Date(data.hackathonStart))
+        setEnd(new Date(data.hackathonEnd))
```
sidyakinian, line 17: Just a small variable naming change, unrelated

PR #554, opened 2024-01-06
Remove waivers

meleongg:
## Description

- fix waivers

### DIFF
--- a/src/components/ApplicationDashboard.js
+++ b/src/components/ApplicationDashboard.js
@@ -337,6 +337,8 @@ const Dashboard = ({
   setReleaseLiabilityCheck,
   mediaConsentCheck,
   setMediaConsentCheck,
+  ageOfMajoritySelect,
+  setAgeOfMajoritySelect,
   username,
   editApplication,
   relevantDates,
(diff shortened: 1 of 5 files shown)


src/utility/Constants.js
```diff
@@ -21,6 +21,15 @@ export const NOTIFICATION_PERMISSIONS = Object.freeze({
   DENIED: 'denied',
 })
 
+export const WAIVER_LINKS = Object.freeze({
+  COVID:
+    'https://docs.google.com/document/d/1td2BgcwRT2CTYdUrzHji7onxpoVl8UjMiEuqIIUK7CM/edit?usp=sharing%5C',
```
EdwardJXLi, line 26: We should remove the trailing space in case that causes issues, but that should be fine

src/pages/Home.js
```diff
@@ -27,7 +24,7 @@ const TopDecal = styled.img`
 `
 
 export default withTheme(({ announcements, theme }) => {
-  const { user, isAuthed } = useAuth()
+  // const { user, isAuthed } = useAuth()
```
EdwardJXLi, line 27: Why are we removing auth?

src/components/ApplicationDashboard.js
```diff
@@ -375,15 +420,99 @@ const Dashboard = ({
...
+
+            <CheckboxContainer>
+              <QuestionLabel>
+                Release of Liability <Required />
+              </QuestionLabel>
+              <P>
+                This waiver allows nwPlus to use any photos or videos taken during the event for
+                promotional purposes.
+              </P>
+              <A bolded color="primary" width="130px" href={WAIVER_LINKS.RELEASE_LIABILITY}>
```
EdwardJXLi, line 440: Very small issue, but we should make the wavers open in the new tab as sometimes the state doesnt get saved when you click directly.

src/pages/Home.js
EdwardJXLi: Okay its for the QR. should be fine for now

EdwardJXLi (approved):
LGTM

byronwang93 (approved):

PR #564, opened 2024-02-07
Updated firebase name fields and changed Dev Portal to cmd-f

meleongg:
## Description

- updated firebase name fields to have legal- prefix
- changed dev portal env to point to cmd-f 2024

## Other considerations

image

- there is a "Hi name" (see code [here](https://github.com/nwplus/portal/blob/preferred-name/src/App.js#L87)) in our Portal navbar, but this currently uses the information from the user's Gmail account and I don't think there is an alternative method to change this to the firebase preferredName field since the user would have to provide their preferred name first, which would mean the navbar would be empty for that time and it would be weird if it kept updating as they changed their preferredName question on the portal application
image

### DIFF
--- a/src/components/ApplicationForm/BasicInfo.js
+++ b/src/components/ApplicationForm/BasicInfo.js
@@ -381,37 +381,37 @@ export default ({ refs, errors, formInputs, onChange }) => (
       </SubHeading>
       <StyledTextInput
         placeholder="First Name"
-        value={formInputs.firstName}
-        errorMsg={errors?.firstName}
-        invalid={!!errors?.firstName}
+        value={formInputs.legalFirstName}
+        errorMsg={errors?.legalFirstName}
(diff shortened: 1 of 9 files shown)


DonaldKLee (changes requested):
questions lgtm, but found an error

after submitting application, i get this
image

DonaldKLee:
> "there is a "Hi name" (see code [here](https://github.com/nwplus/portal/blob/preferred-name/src/App.js#L87)) in our Portal navbar, but this currently uses the information from the user's Gmail account and I don't think there is an alternative method to change this to the firebase preferredName field since the user would have to provide their preferred name first, which would mean the navbar would be empty for that time and it would be weird if it kept updating as they changed their preferredName question on the portal application"

i think having it change as they enter in their preferred name would be a cool feature to have LOL

PR #566, opened 2024-02-10
Escape Node v16

lsha0730:
- Upgraded react-scripts
- Dealt with no anonymous default exports error


meleongg:
Does it work if you remove our node 16 and firebase-tools band-aid fixes to the github ci cd workflow files (.github/workflows)?

lsha0730:
@meleongg Are the bandaid fixes the forced old version numbers in the workflow files? The preview seems to work fine after removing the versions, but lmk if I removed the band-aid properly

meleongg:
Yup, the band-aid fixes are the hardcoded version numbers

meleongg:
We'll prolly merge this one after cmd-f in case there's something bad that happens LOL

meleongg (approved):
Let's hope this doesn't break anything :D

PR #585, opened 2024-06-01
Deploy dev portal for logs to test + fix eslint issue

DonaldKLee:
Git actions was failing due to an eslinting issue with the NotificationsToggle function.
Logs also wanted to run a prev version of portal to see the questions

Merge to main after logs is done with testing?


martincai8 (approved):
🚀

PR #587, opened 2024-06-11
Hackathon context + routing changes

martincai8:
## Description
Documentation: https://www.notion.so/Portal-Documentation-3a2b518a80fa4b9e8dca2d788267730f?pvs=4
- Created a hackathon context to keep track of the active hackathon (one of `hackcamp`, `nwhacks`, and `cmd-f` -> yes ik they're lowercase :angrycry:) as well as the corresponding firebase collection's name
- Whenever the user navigates to a page with a /app/:hackathon/ url, the active hackathon gets set in context + local storage (so that it persists through refreshes) -> this gets cleared when the user navigates back to the home page
- The current theme is also automatically set based on the `activeHackathon` value and it defaults to `nwhacks` to prevent the app going bananas when there's no active theme (since a lot of the components read from the theme)
- Also fixed up some small misc. bugs I ran into while testing but I can move them into another PR if this PR is too big

## Testing
- Change [this](https://github.com/nwplus/portal/blob/ee8e9eb2206445b582d4b27a4a76209333960e4d/src/utility/firebase.js#L43-L47) to the code snippet below to toggle between the application status while testing since the app redirects users to `/app/:hackathon/application` when applications are open
```
export const getLivesiteDoc = callback => {
  return livesiteDocRef.onSnapshot(doc => {
    const data = doc.data()
    data.applicationsOpen = false (or true)
    callback(data)
  })
}
```
- go ham with navigating through the different pages (but not too ham), some of them require admin perms (use your nw email)

### DIFF
--- a/src/components/Navbar.jsx
+++ b/src/components/Navbar.jsx
@@ -4,6 +4,7 @@ import cmdfIcon from '../assets/cmdf_logo.png'
 import nwplus_logo from '../assets/nwplus_icon.svg'
 import { Button } from './Input'
 import { P } from './Typography'
+import { useHackathon } from '../utility/HackathonProvider'
 
 const NavContainer = styled.div`
   direction: rtl;
@@ -54,7 +55,9 @@ const Wrapper = styled.div`
   padding-top: 1rem;
 `
 
-const NavBar = ({ name, handleLogout, children, theme }) => {
+const NavBar = ({ name, handleLogout, children }) => {
+  const { activeHackathon } = useHackathon()
+
   return (
     <Wrapper>
       <NavContainer>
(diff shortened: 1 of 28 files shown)


martincai8:
@tdanielles @DonaldKLee still working on the documentation but feel free to start reviewing first 🫡

src/components/Navbar.jsx
```diff
@@ -71,16 +74,16 @@ const NavBar = ({ name, handleLogout, children, theme }) => {
             <Greeting>Hi, {name}</Greeting>
           </>
         )}
-        {theme.name === 'nwHacks' && (
+        {activeHackathon === 'nwhacks' && (
```
tdanielles, line 77: I may be missing something, but is there a reason why there is no case for hackcamp LOL

src/components/Sidebar.jsx
```diff
@@ -30,14 +30,14 @@ const SidebarContainer = styled.div`
...
+    case 'hackcamp':
       return hc_logo
-    case 'cmdf':
+    case 'cmd-f':
       return cmdf_logo
-    case 'nwHacks':
-      return nwplus_logo
-    default:
+    case 'nwhacks':
       return logo
```
tdanielles, line 38: Why don't we use the nwhacks_logo.svg? or if we want to use logo.svg because I see it is a brighter version would it be better to change the filename to nwhacks_logo.svg for consistency? Not sure if hackathon logos change every year- if it does, ignore everything I just wrote LOL

tdanielles (commented):
Also, not sure if it's just me, but when I go from home to any hackathon, it glitches to show the graphic for `/app/:hackathon` for like a sec or two before navigating to `/app/:hackathon/application`

src/components/Navbar.jsx
martincai8: you're so right

src/components/Sidebar.jsx
martincai8: turns out the nwHacks logo in logo.svg is super old LOL i just changed it to use the updated one, we can remove the outdated logos in our refactor

tdanielles (approved):
lgtm :D

byronwang93:
cracked dev @martincai8

PR #589, opened 2024-07-09
Fix white screen while refreshing application

martincai8:
## Description
- Set the default state for the application to be `null` instead of `HACKER_APPLICATION_TEMPLATE` to better communicate that the application hasn't been loaded yet (versus just an empty application)
- Added a loading state to stop the rendering of any of the children components until `application` has been loaded from firebase
I think the reason why it's broken rn is because it now takes slightly longer to load the application from firebase, and causing the context provider to prematurely give `Part0.jsx` the default value of `HACKER_APPLICATION_TEMPLATE` (which is a blank application) instead of the actual application from firebase. Instead, the value for the application should be `null` until the application is loaded.

## Testing
dev branch
- run the current dev branch (or go to the [preview link](https://dev-nwplus-portal.web.app/))
- select cmd-f, and click complete a registration (you should be brought to the part 0 page)
- reload, there should be a white screen with some error messages in the console saying `TypeError: u.toDate is not a function`

this branch
- do the same on this branch (or using the preview link for this PR)
- you should be able to refresh the part 0 page without any errors
- edit your application/go through the application process and double check that everything still updates in firebase (just in case LOL)

### DIFF
--- a/src/utility/HackerApplicationContext.jsx
+++ b/src/utility/HackerApplicationContext.jsx
@@ -42,7 +42,7 @@ export const uploadWaiverToStorage = async (userId, file) => {
 
 export function HackerApplicationProvider({ children }) {
   const { user } = useAuth()
-  const [application, setApplication] = useState(initialApplication)
+  const [application, setApplication] = useState(null)
   const [, setUpdated] = useState(false)
   const [applicationOpen, setApplicationOpen] = useState(null)
   const applicationRef = useRef()
--- a/src/utility/HackerApplicationContext.jsx
+++ b/src/utility/HackerApplicationContext.jsx
@@ -163,7 +163,7 @@ export function HackerApplicationProvider({ children }) {
     })
   }, [])
 
-  if (isLoading || applicationOpen === null) {
+  if (isLoading || applicationOpen === null || application === null) {
     return null
   }
 
--- a/src/utility/HackerApplicationContext.jsx
+++ b/src/utility/HackerApplicationContext.jsx
@@ -42,20 +42,26 @@ export const uploadWaiverToStorage = async (userId, file) => {
 
 export function HackerApplicationProvider({ children }) {
   const { user } = useAuth()
-  const [application, setApplication] = useState(HACKER_APPLICATION_TEMPLATE)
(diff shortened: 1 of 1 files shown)


tdanielles (approved):
lgtm! goated

DonaldKLee:
What do you mean by:
"Set the default state for the application to be null instead of HACKER_APPLICATION_TEMPLATE to better communicate that the application hasn't been loaded yet (versus just an empty application)"

Do you have an example of what is different on firebase?

I'm getting a white screen when refreshing "https://dev-nwplus-portal.web.app/app/cmd-f/application/part-0"

note: this is from the preview url

DonaldKLee:
for localhost, im getting this for: http://127.0.0.1:3000/app/cmd-f/application

image

src/utility/HackerApplicationContext.jsx
DonaldKLee, line 1: by setting their application to null, would we lose the default fields that firebase add to a user's document upon logging into portal? if so, how can we efficiently update hacker statuses + edit other fields for walk-ins on the day of hackathons?

martincai8:
@DonaldKLee the way it is right now, the hacker application context gets initialized with the `application` state = a blank application -> fetch the hacker's actual application from firebase -> fill in any missing values in the actual application -> update `application`. logically, i think it would make more sense if it is initialized with null (because the actual application hasn't been fetched yet from firebase), and then set `application` once everything is ready. 

if you try to run dev locally, and put a `console.log(application)` in `Part0.jsx` and refresh, you'll see that `application` is equal to a blank application, which isn't the best representation of the application because the hacker's application isn't actually blank, it just hasn't been fetched from firebase yet (this also breaks the application form because it assumes that the application should have a last updated by field, which is blank in the blank application)

DonaldKLee (changes requested):
check question about setting application to null

loading code lgtm
q: do we know why pulling the application takes longer now?

martincai8:
firebase: no, firebase isn't actually affected, this application is the one stored in the context provider

@tdanielles is it because we're relying on dbHackathonName now?

PR #593, opened 2024-09-10
new landing page

martincai8:
## Description

figma link: https://www.figma.com/design/bSh3usa74gfv1VvHH7Z4bf/nwPlus-Portal-Live-Site?node-id=3507-2680&node-type=vector&t=Vofx4QkFSQ3kK87x-0

- The hackathon dates, applicationOpen statuses, application deadlines are grabbed from firebase. pls feel free to change the values in the dev firebase and test that they change the corresponding values
- `visitWebsite` is a prop that's manually passed into a card that makes the button link to a hackathon's website instead of portal (used when everything's done or if we don't wanna link to portal for whatever reason)
- i hardcoded `UP_NEXT_HACKATHON_NAME` to be a constant in the code just because I wanted to add the comment explaining the possible values, but i can move this to firebase if that's better!
- decided to not do the end page since it complicates the logic unnecessarily LOL (but i can also do it if needed)

Desktop: 
image

Mobile:
| up next hackathon | preview |
|----------------|-------|
| HackCamp | image |
| nwHacks | image |
| cmd-f | image |


src/pages/HackathonSelection.jsx
```diff
@@ -0,0 +1,410 @@
...
+
+  const websites = {
+    'HackCamp': 'https://hackcamp.nwplus.io',
+    'nwHacks': 'https://nwhacks.io',
+    'cmd-f': 'https://cmd-f.nwplus.io',
+  }
+  window.open(websites[hackathonName], '_blank', 'noopener,noreferrer')
+}
+
+const HackathonCard = ({
```
tdanielles, line 271: would it be better to move this to another file under components along with the related css / styling code above just so this file isn't too cluttered

tdanielles (changes requested):
lgtm overall! so cracked

joannelex:
legend

joannelex (approved):

src/pages/HackathonSelection.jsx
martincai8: 💯 🫡 ✅

tdanielles (approved):

PR #594, opened 2024-09-23
remove MLH checkboxes for HC

martincai8:
## Description
- removed all the MLH checkboxes if the active hackathon is HackCamp (except the code of conduct)
- only ask the gender acknowledge question if the active hackathon is cmd-f


DonaldKLee (approved):
lgtm

PR #596, opened 2024-09-24
some bug fixes

martincai8:
## Description
fixed:
- long answer word count validation + error messages
- judging redirect button was missing a slash

### DIFF
--- a/src/components/ApplicationQuestions/LongAnswer.jsx
+++ b/src/components/ApplicationQuestions/LongAnswer.jsx
@@ -9,7 +9,6 @@ const StyledTextArea = styled(TextArea)`
 const LongAnswer = ({ refs, errors, formInputs, onChange, question }) => {
   return (
     <>
-      {errors?.[question.formInput] && <ErrorMessage>{errors?.[question.formInput]}</ErrorMessage>}
       <StyledTextArea
         maxWords={question.maxWords || '150'}
         width="100%"
--- a/src/components/Input/TextArea.jsx
+++ b/src/components/Input/TextArea.jsx
@@ -43,17 +43,11 @@ export const TextArea = ({
     }
     return cleanedSplit.length || 0
   }, [value])
-  useEffect(() => {
-    // debounced with setTimeout
-    const handler = setTimeout(() => {
-      setIsLengthExceeded(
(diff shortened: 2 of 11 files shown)


src/components/Input/TextArea.jsx
```diff
@@ -43,17 +43,11 @@ export const TextArea = ({
...
-      )
-    }, 500)
 
-    return () => {
-      clearTimeout(handler)
-    }
+  useEffect(() => {
+    setIsLengthExceeded(
+      maxLength ? value && value.length > maxLength : maxWords ? getWords() > maxWords : false
+    )
```
martincai8, line 50: got rid of the debounce because we're conditionally showing error messages. the debounce caused a delay -> resulting in the `isLengthExceeded` error message to flash when we go from over word count to under word count

src/components/ApplicationQuestions/LongAnswer.jsx
```diff
@@ -9,7 +9,6 @@ const StyledTextArea = styled(TextArea)`
 const LongAnswer = ({ refs, errors, formInputs, onChange, question }) => {
   return (
     <>
-      {errors?.[question.formInput] && <ErrorMessage>{errors?.[question.formInput]}</ErrorMessage>}
```
martincai8, line 12: the error message is already shown in the `TextArea` component

src/components/Input/TextArea.jsx
```diff
@@ -73,17 +67,19 @@ export const TextArea = ({
...
-            ? `${value.length - maxLength}  character(s)`
+            ? `${value.length - maxLength} character(s)`
             : `${getWords() - maxWords} word(s)`}{' '}
-          over the limit.{' '}
+          over the limit.
         </ErrorMessage>
+      ) : (
+        invalid && <ErrorMessage>{errorMsg}</ErrorMessage>
       )}
+
```
martincai8, line 82: these two error messages are basically the same, but kept them both in to be safe. when the length is exceeded, the more descriptive error message is shown, and then when the response is invalid (due to validation for hacker app + other factors in other areas), then it shows the error message that was passed in

tdanielles (approved):

PR #599, opened 2024-09-29
Portal theming variables + HC reskin

martincai8:
## Description
finally lol

## Other considerations
- set nwHacks2024 as the database name so events will show up, will change it back to nwHacks2025 before merging


src/components/Sidebar.jsx
```diff
@@ -318,7 +308,7 @@ const Sidebar = ({
                 )
               )
             })}
-            {!IS_DEVICE_IOS ? <NotificationToggle /> : null}
+            {/* {!IS_DEVICE_IOS ? <NotificationToggle /> : null} */}
```
joannelex, line 311: do we need this LOL

src/pages/Home.jsx
```diff
@@ -52,10 +62,6 @@ const HomeContainerBackground = styled.div`
 //   width: calc(100vw - 256px + 2em);
```
joannelex, line 62: comment?

src/pages/Home.jsx
```diff
@@ -8,19 +8,20 @@ import QrCode from '../components/QrCode'
 import { useAuth } from '../utility/Auth'
 // import Hackcamp2023BG from '../components/BackgroundImage'
```
joannelex, line 9: i assume we re-use this ?

src/components/Sidebar.jsx
joannelex: so leave it as comment? can we just remove it?

joannelex (approved):
goat

src/utility/Constants.js
```diff
@@ -2,7 +2,7 @@ export const DB_COLLECTION = 'Hackathons'
 
 export const DB_HACKATHON_NAMES = {
   'hackcamp': 'HackCamp2024',
-  'nwhacks': 'nwHacks2025',
+  'nwhacks': 'nwHacks2024',
```
DonaldKLee, line 5: should this be nwHacks2025 if cmd-f is using 2025?

DonaldKLee (approved):
lgtm, just left 1 comment

src/components/Sidebar.jsx
martincai8: i kept it in because most of the logic is built already, so if we decide to use it in the future then we won't have to rebuild it from scratch

PR #624, opened 2024-10-22
HC RSVP Page Update

tdanielles:
## Description
- updated waiver links for hackcamp
- added HC feature question for marketing
- removed mentorship question


martincai8 (dismissed):
lgtm!! ty for taking it on 🙇‍♂️ can you take a look at my commit to make sure it looks okay too

let's keep this open and make sure HC is okay with it before merging it in

PR #627, opened 2024-10-24
Attended Events component

daniel-panhead:
## Description

image

slay bro

Added attended events component! Events with no type are just gray, otherwise they correspond to the legend. 

## Other considerations

Also added some basic scaffolding for the events page, feel free to use or discard for laying out other components.

Overflow of the components was added but currently doesn't work since the parent container doesn't have a max height. Once other components are added we can figure out a reasonable max height for the component ^^


ErpingS (approved):
lgtm

PR #629, opened 2024-10-29
Wrap getAndAssignProjects with db.runTransaction

tdanielles:
## Description
- wrap `getAndAssignProjects` with `db.runTransaction` as an attempt to fix a potential race condition that happens when projects are being assigned


martincai8 (approved):
lgtm! would it be possible this in a try catch so we can debug on the day of if needed 😰

PR #630, opened 2024-10-31
Rewards component UI

indirasowy:
## Description
Added rewards card component to the Rewards page. Rewards are fetched from Firebase

Screenshot 2024-10-31 at 12 03 21 PM
Screenshot 2024-10-31 at 12 10 29 PM

## Other considerations
Code that displays points associated with each rewards + card flip has been added and commented out as points will not be used for Hackcamp. I left it in as commented code as it will be used in the future for nwHacks.


DonaldKLee (approved):
LGTM, great work!

- Removed the container for user stats since it wont be used for hackcamp
- added column gaps based off figma

PR #634, opened 2024-11-01
Dev -> Main

DonaldKLee:


src/components/RewardCard.jsx
```diff
@@ -0,0 +1,163 @@
...
+  color: ${p => p.theme.colors.lines};
+`
+
+const SubTitle = styled.p`
+  font-size: 14px;
+  font-weight: 700;
+  color: ${p => p.theme.colors.bar};
+  text-align: right;
+  margin-top: 8px;
+`
```
martincai8, line 48: should these colours be `theme.colors.text`?

src/components/RewardCard.jsx
```diff
@@ -0,0 +1,163 @@
...
+
+const InfoIcon = styled.div`
+  position: absolute;
+  top: 16px;
+  right: 16px;
+  width: 24px;
+  height: 24px;
+  border: 2px solid ${p => p.theme.colors.lines};
+  border-radius: 50%;
+  color: ${p => p.theme.colors.lines};
```
martincai8, line 64: this one too!

src/components/RewardCard.jsx
```diff
@@ -0,0 +1,163 @@
...
+`
+
+const ProgressBar = styled.progress`
+  -webkit-appearance: none;
+  appearance: none;
+  width: 100%;
+  height: 20px;
+
+  &::-webkit-progress-bar {
+    background-color: ${p => p.theme.colors.barBackground};
```
martincai8, line 91: i think design forgot to add variables for these `barBackground` and `bar`, please let them know to add it so everything's synced up

martincai8 (changes requested):
lgtm! just a couple of colour variable things to watch out for

src/components/RewardCard.jsx
martincai8: oh idk why it didn't highlight correctly, but i was talking about all the text colours

indirasowy: it should be the same color as the bar! it's for the points here above the bar image

martincai8: ohhh okay yeah that sounds good! should the the other ones should be `colors.text` instead of `colors.lines`?

src/components/RewardCard.jsx
indirasowy: oh it looks like theme.colors.lines and theme.colors.text both have the same hexcode of #45171A, should I just change it to text so it's more consistent or is it fine to leave it?

martincai8: yeah i see why that was confusing LOL 

it would be amazing if you could change it to text so it matches up with the variable being used in figma! in case those two variables are different for future reskins

martincai8: i'll write the docs for this sometime it's my bad for not specifying it sorry

indirasowy: ok sounds good! if i push another change do you know if i have to open a new PR because this one was merged?

martincai8: yepp i can help merge it in

PR #643, opened 2024-11-13
Allow admins to view application pages anytime

martincai8:
## Description

Allows admins to access the hacker application pages through `/app/{activeHackathon}/application/page-x` even when apps are not open yet so they can view the app questions


DonaldKLee (approved):
so fast

PR #659, opened 2024-12-29
Update rewards page for nwhacks

indirasowy:
## Description
- Updated rewards page to include points logic
- Use DayOf collection instead of Events

lmk if I missed anything!

Screenshot 2024-12-28 at 4 31 32 PM

### DIFF
--- a/src/containers/Rewards.jsx
+++ b/src/containers/Rewards.jsx
@@ -11,7 +11,6 @@ import { useAuth } from '../utility/Auth'
 const Container = styled.div`
   display: grid;
   grid-template-columns: 1.2fr 1.8fr;
-  align-items: center;
 
   @media (max-width: 768px) {
     display: flex;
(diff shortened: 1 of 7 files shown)


src/containers/Rewards.jsx
```diff
@@ -61,12 +83,37 @@ const Rewards = () => {
...
+    const getUserPoints = async () => {
+      if (!user) return
+      try {
+        const userDoc = await applicantsRef(dbHackathonName).doc(user.uid).get()
+        if (!userDoc.exists) {
+          console.log('No user data found')
+          return
+        }
+        const userData = userDoc.data()
+        const points = userData.points || 0
```
DonaldKLee, line 95: points are calculated from taking the events the hacker has attended from dayOf (dayOne, dayTwo and other) and then cross checking the actual event in dayOf and getting the points value -> total up all points

DonaldKLee (changes requested):

PR #670, opened 2025-01-10
portal rewards - 

byronwang93:
## Description
- on portal -> rewards, added an extra field to specify which type each prize is
ex of a prize w/ no type field (same as original)
Screen Shot 2025-01-09 at 5 43 33 PM

ex of a prize w/ a type field (newly added)
Screen Shot 2025-01-09 at 5 43 42 PM

### DIFF
--- a/src/components/RewardCard.jsx
+++ b/src/components/RewardCard.jsx
@@ -136,23 +136,25 @@ const RewardCard = ({ name, type, desc, company, image, points, requiredPoints }
             ) : (
               <Text>Reach {requiredPoints - points} pts</Text>
             )}
-            {type && <Text>Type: {type}</Text>}
+            {type === 'Raffle' ? <Text>Raffle Prize</Text> : type && <Text>Type: {type}</Text>}
           </div>
         </Grid>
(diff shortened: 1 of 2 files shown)


src/components/RewardCard.jsx
```diff
@@ -136,6 +136,7 @@ const RewardCard = ({ name, desc, company, image, points, requiredPoints }) => {
             ) : (
               <Text>Reach {requiredPoints - points} pts</Text>
             )}
+            {type && <Text>Type: {type}</Text>}
```
DonaldKLee, line 139: can u do if points exist, display points bar, if type is raffle, say "Raffle Prize" and else, display "Type: [type}"

DonaldKLee (changes requested):

src/components/RewardCard.jsx
byronwang93: done, also made it so if type doesn't exist, then the `Type: [type]` doesn't show up

kashgarg (approved):
lgtm - donald

kashgarg (approved):

PR #676, opened 2025-01-17
Add overlay for reward if it has already been redeemed

daniel-panhead:
## Description

Quick change to indicate to users when a certain reward has already been redeemed by them

image


DonaldKLee (approved):
lgtm

PR #692, opened 2025-02-19
change the data in the qr code to be a link

martincai8:
## Description
Stores the data in the qr code as `{baseUrl (portal.nwplus.io or localhost or dev-nwplus...)}/app/{activeHackathon}/socials/{userId}` instead of just the userId

QR PR: https://github.com/nwplus/qr-check-in/pull/136


byronwang93 (approved):
cracked as always

PR #694, opened 2025-02-22
build socials page for desktop

martincai8:
## Description
I apologize for the messy CSS LOL I'll clean it up when I work on mobile, but this should set up all of the relevant data fetching and setting (except recently viewed profiles) so everyone can start working on the other parts.

Logic is the same as explained in the slack thread, but basically we'll only grab preferred name and pronouns from `Applicants` if `Socials` doesn't have it, otherwise all data should be read from `Socials`. The page can be viewed when the user is not logged in now too.

link to dummy profile: https://nwplus-ubc--pr694-build-socials-page-wq2tv2ls.web.app/app/cmd-f/social/xGRhSFg0z3NDom6dZgtKDmVr76h1
2025-02-22 at 12 31PM


src/components/SocialIcons.jsx
```diff
@@ -0,0 +1,47 @@
...
+        <a key={platform} href={link} target="_blank" rel="noopener noreferrer">
+          {isCustom ? (
+            <IconComponent color={theme.colors.text} size="2em" />
+          ) : (
+            <Icon icon={IconComponent} brand={brand} size="2x" color={theme.colors.text} />
+          )}
+        </a>
+      )
+    })}
+  </SocialIconsContainer>
```
byronwang93, line 44: not sure if this is in the scope of the ticket but if the url doesn't have `https://` then it just appends the url to the main url

should we create helper to automatically add https if it doesnt exist in url?

ex clicking on your devpost url -> `https://nwplus-ubc--pr694-build-socials-page-wq2tv2ls.web.app/app/cmd-f/social/devpost.com`

martincai8: yess good point, would you be able to add link validation to your PR (when editing their info)? e.g., should check that the links are valid (github, devpost, instagram, etc.) before saving them

byronwang93: oooo wait yes i can add that to my edit functionality

byronwang93 (approved):

PR #698, opened 2025-02-25
base edit functionality completed

byronwang93:
## Description

- created the edit profile page
- separated the existing `Social.jsx` into `EditSocial.jsx` and `ViewSocial.jsx`
- data correctly saves to firebase
- save button disabled if name or pronouns empty
- created new constants for roles, and year level
- created `ensureHttps` function to ensure urls have `https` in front (or else firebase will just append url onto existing base url)

## Other considerations

- mobile view not created yet, will make separate ticket

### DIFF
--- a/src/components/Social/EditSocial.jsx
+++ b/src/components/Social/EditSocial.jsx
@@ -4,6 +4,9 @@ import { Button, Checkbox, Dropdown, TextArea, TextInput } from '../Input'
 import veebs from '../../assets/profilePictures/veebs.svg'
 import { ensureHttps } from '../../utility/utilities'
 import schoolOptions from '../../containers/Application/data/schools.json'
+import majors from '../../containers/Application/data/majors.json'
+import roles from '../../containers/Application/data/roles.json'
+import years from '../../containers/Application/data/years.json'
 
 const TopRow = styled.div`
   display: flex;
@@ -50,6 +53,7 @@ const InputContainer = styled.div`
 const InputDropdown = styled(Dropdown)`
   .react-select__control {
     height: 2.9rem;
+    max-width: 20.3rem;
     min-height: 2.5rem;
   }
 `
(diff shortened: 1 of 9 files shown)


src/containers/Social.jsx
```diff
@@ -190,6 +103,15 @@ const Social = ({ userId }) => {
...
   }, [userId, user?.uid])
 
+  const saveUserData = async updatedData => {
+    try {
+      await socialsRef.doc(currentUserId).set(updatedData, { merge: true })
+      console.log('saved!')
+    } catch (error) {
+      console.error('error saving user data: ', error)
+    }
+  }
```
martincai8, line 113: this is smart could you use this in line 165 too (replace `await socialsRef.doc(currentUserId).set(finalData, { merge: true })`)

martincai8 (approved):
lgtm overall you're so cracked 🐐

noticed a small shift when changing from `ViewSocial` to `EditSocial`, maybe something to do with margins/paddings?

https://github.com/user-attachments/assets/f22dc247-0296-444e-beaa-1e713c7a633b

src/containers/Social.jsx
martincai8: oops meant to highlight `saveUserData`

byronwang93: cracked feedback

src/components/Social/EditSocial.jsx
```diff
@@ -0,0 +1,270 @@
...
+              color="primary"
+              width="flex"
+            >
+              Discard Changes
+            </EditProfileButton>
+            <EditProfileButton
+              disabled={name.trim() === '' || newPronouns === ''}
+              onClick={() => {
+                if (name.trim() !== '' && newPronouns !== '') {
+                  setIsEditing(false)
```
byronwang93, line 152: maybe something to look into later but our `Button` component is a styled `a` instead of `button` ?? so the disabled state doesn't inherently block the onClick functionality -> so we need a check in the onClick too oop

byronwang93:
@martincai8 comments (hopefully) addressed now :P

src/containers/Application/data/years.json
```diff
@@ -0,0 +1,30 @@
+[
+  {
+    "id": 1,
+    "label": "Pre-University"
+  },
+  {
+    "id": 2,
```
byronwang93, line 7: taken from admin

martincai8 (approved):
CRACKED THIS IS SO GOOD

PR #702, opened 2025-03-03
Added Recently Viewed profiles section

indirasowy:
## Description
- Added Recently Viewed profile section to a user's profile with hyperlinks to navigate back to a listed profile
- Added hide recently viewed logic for users who opt out of appearing on other people's Recently Viewed
- Added mobile styling
Screenshot 2025-03-02 at 1 26 56 PM


src/assets/delete.svg
martincai8, line 1: i think we might need to pass colours into the svgs so it can take on be reskinned properly! like how it's done here: https://github.com/nwplus/portal/blob/3e6156f07405856adba8c50165791ed59500db7c/src/components/Schedule/Tag.jsx#L41-L47

the normal one uses `p.theme.colors.text` and the hover uses `p.theme.colors.button.primary.text`. you can also just remove the hover svg after since you can adjust the fill with a prop

src/components/Social/ViewSocial.jsx
```diff
@@ -308,8 +417,29 @@ const ViewSocial = ({
...
-        {activeTab === 'Recently Viewed' && (
-          <p>Recently Viewed Profiles</p> // placeholder
+        {isCurrentUser && (isMobile ? activeTab === 'Recently Viewed' : true) && (
+          <>
+            <HeaderText>Recently Viewed Profiles</HeaderText>
+            {recentlyViewedProfiles.length > 0 ? (
+              <RecentlyViewedContainer>
+                {recentlyViewedProfiles.map((profile, index) => (
+                  <Profile key={index}>
+                    <Username href={`/app/cmd-f/social/${profile.profileId}`}>
```
martincai8, line 427: would it be possible to update this to `/app/{activeHackathon}/social/${profile.profileId}` so it can be dynamic across hackathons (since this feature won't be specific to cmd-f)? you can get `activeHackathon` by doing
- import { useHackathon } from '../../utility/HackathonProvider'
- const { activeHackathon } = useHackathon()

example from the QR code: https://github.com/nwplus/portal/blob/3e6156f07405856adba8c50165791ed59500db7c/src/components/QrCode.jsx#L242

src/components/Social/ViewSocial.jsx
```diff
@@ -228,6 +230,103 @@ const MobileSocialIconsContainer = styled.div`
   }
 `
 
+const HeaderText = styled.h1`
+  font-size: 32px;
+  font-weight: 800;
+  margin-top: 1rem;
+
+  ${p => p.theme.mediaQueries.mobile} {
+    font-size: 28px;
```
martincai8, line 239: i think the header is left aligned on mobile!
```suggestion
    font-size: 28px;
    align-self: flex-start;
    padding-left: 5%;
```

| Figma                                                        | After                                                         |
| :------------------------------------------------------------ | :------------------------------------------------------------ |
| 2025-03-02 at 10 40PM | 2025-03-02 at 10 41PM |

src/components/Social/ViewSocial.jsx
```diff
@@ -228,6 +230,103 @@ const MobileSocialIconsContainer = styled.div`
...
+  max-height: 300px;
+  overflow-y: auto;
+  margin-top: -24px;
+
+  ${p => p.theme.mediaQueries.mobile} {
+    max-height: none;
+    overflow-y: visible;
+    justify-content: center;
+    align-items: center;
+    margin-top: -18px;
```
martincai8, line 256:
```suggestion
    margin-top: -18px;
    width: 100%;
```

to make it a bit less squished 🤏

| Before                                                        | After                                                         |
| :------------------------------------------------------------ | :------------------------------------------------------------ |
| 2025-03-02 at 10 42PM | 2025-03-02 at 10 42PM |

martincai8 (changes requested):
THIS IS SO COOL!! logic lgtm 🚀

left some comments about small things, and maybe it's just me but the trash icon doesn't show up on desktop, but otherwse gjgjgj
2025-03-02 at 10 31PM

src/components/Social/ViewSocial.jsx
indirasowy: Oh good point, will do that!

martincai8:
@indirasowy could you add the rest of the fixes in a new PR 🙇‍♂️

indirasowy:
@martincai8 Oh yes sorry I couldn't get to it, will open a new PR by today!

PR #706, opened 2025-03-04
Jason default points change

jkuo630:
image

- set default points to 15 on frontend and backend 
- also made a safeguard such that if points fall to 0, it shouldn't reset back to default; it should stay at 0


DonaldKLee:
Might handle this via QR app instead, otherwise they'd need to view the rewards page to get their points, mb if the ticket wasn't clear

i can make a pr for this rn

jkuo630:
o shi okok mbmb

DonaldKLee:
wait nvm im dumb, this is good LOL, cause we don't have a field to keep track of points 😂

DonaldKLee:
gonna push a commit then merge!

jkuo630:
LOL okok sound goood thanks!!

PR #713, opened 2025-03-06
quick fix done to make profile picture in edit mode a circle, rather than a circle 

jkuo630:
new 
Screenshot 2025-03-05 at 4 37 11 PM
old
Screenshot 2025-03-05 at 4 37 42 PM


martincai8 (approved):
👑

PR #718, opened 2025-05-21
Set up next on the homepage to null

martincai8:
## Description

- added `visitWebsite` to cmd-f
- set up next on the homepage to null

Before
2025-05-20 at 08 17PM

After
image


Khaedra (approved):
very awesome

PR #720, opened 2025-07-23
save to images/download functionality

Khaedra:
## Description

Download (desktop) and Save to Images (mobile) functionality for QR codes. 

## Other considerations

New Desktop View
image

New Mobile View
image

Downloaded Image preview
image

How it Works: 
1. Reformatted QR Ticket DOM structure to be simpler
2. Generate canvas from ticket, convert to blob
3. If mobile, use default share functionality 
4. If errors occur or are on a desktop, creates a link to the image and automatically downloads. 

User Exp:
- Mobile: Press "download image", opens up default sharing popup, can save to images, downloads, share in text etc.
- Desktop: press button, yay image downloaded. 

Problems/Caveats:
- nwHacks background svg doesn't specify height/width, making it appear super zoomed in on download. Hackcamp and cmd-f background/downloads are fine. For future hackathon background images, specify height and width. 

- Deleted "visitWebsite" attribute for HackCamp so you can visit portal and see your QR Code. Need to change your application status to "acceptedAndAttending" in firebase (Hackathons -> HackCamp2024 -> Applicants -> Your application).

### DIFF
--- a/src/components/QrCode.jsx
+++ b/src/components/QrCode.jsx
@@ -1,6 +1,7 @@
-import React from 'react'
+import React, { useState, useRef } from 'react'
 import styled from 'styled-components'
 import { useQRCode } from 'next-qrcode'
+import html2canvas from 'html2canvas'
 import hcQrBackground from '../assets/hc2024qrcode.svg'
 import nwHacksQrBackground from '../assets/nwhacks2025qrcode.svg'
 import cmdfQrBackground from '../assets/cmdf2025qrcode.svg'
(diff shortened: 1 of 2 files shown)


src/components/QrCode.jsx
```diff
@@ -184,6 +178,60 @@ const AppleWalletButton = styled.button`
...
+  font-weight: 600;
+  cursor: pointer;
+  margin-top: 20px;
+  transition: all 0.2s ease;
+  display: flex;
+  align-items: center;
+  gap: 8px;
+
+  &:hover {
+    background: ${p => p.theme.colors.primaryHover || p.theme.colors.primary};
```
martincai8, line 197: i don't think these colours exist in [ThemeProvider.jsx](https://github.com/nwplus/portal/blob/dev/src/theme/ThemeProvider.jsx) (that's why they're not being applied). not a big deal at all, just in case you missed it

src/components/QrCode.jsx
```diff
@@ -215,6 +268,106 @@ const QrCode = ({ userInfo, userId }) => {
...
+    )
+  }
+
+  const generateFilename = () => {
+    const hackathonNames = {
+      'hackcamp': 'HackCamp',
+      'nwhacks': 'nwHacks',
+      'cmd-f': 'cmd-f',
+    }
+    const hackathonName = hackathonNames[activeHackathon] || 'Event'
```
martincai8, line 287: this can be simplified with `const hackathonName = copyText[activeHackathon].hackathonNameShort` 🚀

src/components/QrCode.jsx
martincai8: same with the `background` prop used in the default state

martincai8 (approved):
lgtm to me too! left some small comments but thank you for this it's a much needed feature LOL

